PR #17, opened 2019-10-04
BoxLayout improvements

bylaws:
- Support horizontal scrolling.
- Only fit fill the layout if the horizontal size isn't set, otherwise use that.


natinusala:
Closed as this PR is reimplemented or made irrelevant by the changes on the `yoga` branch. BoxLayout is gone, and scrolling works very differently now.

I will add "horizontal scrolling" to the list of things to implement though.

Issue #30, opened 2020-06-20
Tabframe does not switch to a view if previous view was nullptr.


jon-dez:
Here's an easy fix
```
// File: library/lib/tab_frame.cpp

void TabFrame::switchToView(View* view)
{

    ...
// Line 66
    this->rightPane = view;
    if (this->rightPane != nullptr)
        this->layout->addView(this->rightPane, true, true); // addView() calls willAppear()
}
```

Issue #29, opened 2020-06-20
Segfault when Sidebar has no child


jon-dez:
There is a segfault when borealis tries to render a sidebar with no children.
Unsure if this is intended, but a quick fix is trivial.

```
// File: library/lib/sidebar.cpp

View* Sidebar::getDefaultFocus()
{

    ...

// Line 43
    View* toFocus{ nullptr };
    if(this->children.size() != 0)
        // Try to focus last focused one
        toFocus = this->children[this->lastFocus]->view->getDefaultFocus();

    ...

}
```

PR #31, opened 2020-06-20
Issue #30 and #29, and a suggestion for ease of use.

jon-dez:
I found two bugs (and fixed them of course), one where a program could segfault when a SideBar has no child and where a Tabframe will not switch to any view after it has encountered one view that is a nullptr.
I also modified the constructor of brls::Button to default construct a PLAIN button for ease of use.

### DIFF
--- a/library/lib/sidebar.cpp
+++ b/library/lib/sidebar.cpp
@@ -42,8 +42,8 @@ View* Sidebar::getDefaultFocus()
         this->lastFocus = 0;
 
     View* toFocus{ nullptr };
+    // Try to focus last focused one
     if(this->children.size() != 0)
-        // Try to focus last focused one
         toFocus = this->children[this->lastFocus]->view->getDefaultFocus();
     
     if (toFocus)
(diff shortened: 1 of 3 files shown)


library/lib/sidebar.cpp
```diff
@@ -41,8 +41,11 @@ View* Sidebar::getDefaultFocus()
     if (this->lastFocus >= this->children.size())
         this->lastFocus = 0;
 
-    // Try to focus last focused one
-    View* toFocus = this->children[this->lastFocus]->view->getDefaultFocus();
+    View* toFocus{ nullptr };
+    if(this->children.size() != 0)
+        // Try to focus last focused one
```
natinusala, line 46: Can you move the comment above the if please?

jon-dez: Done!

jon-dez:
No problem!
Just to clarify, the segfault/crash happened on the SideBar when there are no elements in the vector of children.
If the TabBarItem had no view (a nullptr) there was no crash, but it prevented a TabBarItem with a view (not nullptr) from switching if the view it was switching from was nullptr.

PR #32, opened 2020-06-20
Fix macro expansion problems under MSYS2

meganukebmp:
Under MSYS2 on Windows the building using the provided Makefile causes a macro expansion problem (probably due to Windows path slash weirdness). Simply escaping the forward slash seems to fix this

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ APP_VERSION	:=	1.0
 
 ROMFS				:=	resources
 BOREALIS_PATH		:=	.
-BOREALIS_RESOURCES	:=	romfs:/
+BOREALIS_RESOURCES	:=	romfs:\/
 
 #---------------------------------------------------------------------------------
 # options for code generation


natinusala:
Thanks for the PR!

That gives the warning "unknown escaping sequence" when using on Linux. I want to make sure it still runs when built from Linux (since `romfs:\/thing.png` is not a valid path and would probably crash the homebrew).

Another cleaner approach would be to remove that define in the Makefile and put it in the code instead, as a default value:

```
#ifndef BOREALIS_RESOURCES
#define BOREALIS_RESOURCES "romfs:/"
#endif
````

meganukebmp:
I don't like the notion of having to riddle code with includes generally. Configurations imo should be within the config/makefile. Maybe a better solution would be to have a PC/Switch flag. Since the forward slash is only needed for `romfs:/` that can be defined somewhere else and `BOREALIS_RESOURCES` would be ignored. Where as flipping the flag to PC takes `BOREALIS_RESOURCES` into account.

natinusala:
Sorry it was not clear - when I saiy "Linux build" I say build for Switch using Linux as host, as opposed to building for Switch using Windows + MSYS2 as host. The Makefile is only to be used for Switch, actual PC building uses Meson.

meganukebmp:
My mistake, I have completely misread what you said in your first post and rambled out. Yes, that would indeed be cleaner. I assumed manually defining `BOREALIS_RESOURCES` within my own source files and not keeping it as a default. Happy I got this to your attention tho. Cheers!

natinusala:
No worries! Do you want to PR that new fix?

meganukebmp:
Sure just a sec

PR #33, opened 2020-06-20
Set BOREALIS_RESOURCES to default to romfs:/

meganukebmp:
This fixes compilation under Windows paths without breaking unix paths

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,6 @@ APP_VERSION	:=	1.0
 
 ROMFS				:=	resources
 BOREALIS_PATH		:=	.
-BOREALIS_RESOURCES	:=	romfs:/
 
 #---------------------------------------------------------------------------------
 # options for code generation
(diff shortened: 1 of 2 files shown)


natinusala:
You also need to remove the -D from the Switch Makefile

meganukebmp:
I should probably do this type of stuff earlier through the day

natinusala:
There is actually one use case that I had forgotten about, where the borealis resources are in a subfolder of romfs - EdiZon and sys-clk manager do that: `BOREALIS_RESOURCES="romfs:/library/borealis`. That way they can put their own resources in romfs next to the borealis ones.

I would rather not break that feature 😕

meganukebmp:
Tracking this issue down further, I'm about to go insane. Might as well be an issue with MSYS2. Seems that the way it converts unix paths to windows ones and vice versa is some magic text parsing. Which in this event is attempting to flip the backslashes in the romfs escapes into unix forward slashes and making a big mess in the process. Disabling the conversion with `MSYS2_ARG_CONV_EXCL="*"` technically fixes the problem but only if it's being built out of the virtual root. If one tries to build out of the fake links to a windows path it fails miserably.

I still think this needs to be solved somehow. Either on this or on MSYS2. Could potentially be walling off new developers who arent yet ready to jump ship to *nix. I'll keep looking into it.

p-sam:
Yeah this won't work, we already found a solution using `MSYS2_ARG_CONV_EXCL` this morning though. I'll push the PR to disable it for defines only, which seems to be a better solution

meganukebmp:
I figured it out as well. The problematic part is the macro expansion when passing the define to the preprocessor. `-DBOREALIS_RESOURCES="\"$(BOREALIS_RESOURCES)\""` to be specific. The way `MSYS2_ARG_CONV_EXCL` works is that it can ignore values after specified prefixes. This is not a problem with the repository at hand. It's all sound here. It's Microsoft's genius of using the escape character as a path delimiter and the various hacks around all of it.  I'm getting proper builds by exporting 
`MSYS2_ARG_CONV_EXCL="-DBOREALIS_RESOURCES="` before running make. I feel like this should at the very least be documented somewhere. No need to change the repo, it's a platform specific problem. I'll do a PR on the readme.

PR #38, opened 2020-06-26
Progress towards eventual touch support.

jon-dez:
Supporting a mouse could in turn translate to touch support. I don't have a development environment setup to test anything on the switch as of now... but reading touch screen input from the switch and interpreting it as a mouse, with regards to this pull request, should in theory allow for some touch support. There isn't full touch/mouse support yet, but I made some progress and that is what matters right?! :smile:

### DIFF
--- a/library/include/borealis/view.hpp
+++ b/library/include/borealis/view.hpp
@@ -81,7 +81,7 @@ class View
 {
   private:
     Background background = Background::NONE;
-
+    
     void drawBackground(NVGcontext* vg, FrameContext* ctx, Style* style);
     void drawHighlight(NVGcontext* vg, ThemeValues* theme, float alpha, Style* style, bool background);
 
@@ -178,6 +178,13 @@ class View
     }
 
   public:
+    inline bool withinBoundaries(float xpos, float ypos) {
+        return !(
+          (xpos < x || xpos > (x + width))
+          ||
+          (ypos < y || ypos > (y + height))
+        );
+    }
     void setBoundaries(int x, int y, unsigned width, unsigned height);
 
     void setBackground(Background background);
@@ -350,6 +357,29 @@ class View
     {
         return nullptr;
     }
+
+    /**
+     * Return the child at xpos and ypos (in screen coordinated), or null if
+     * there is none at the position.
+     */
+    virtual View* getChildViewAtTouch(float xpos, float ypos);
+
+    /**
+     * Self explanatory.
+     */
+    virtual bool isFocusable() const
+    { return false; }
+
+    /**
+     * Self explanatory.
+     */
+    virtual bool isDraggable() const
+    { return false; }
+
+    /**
+     * Drag a view with a drag delta of dx and dy.
+     */
+    virtual void dragView(float dx, float dy);
 
     /**
       * Fired when focus is gained
--- a/library/include/borealis/view.hpp
+++ b/library/include/borealis/view.hpp
@@ -121,6 +121,7 @@ class View
     bool focused = false;
 
     View* parent = nullptr;
+    std::vector<View*> children;
 
     GenericEvent focusEvent;
 
@@ -203,6 +204,8 @@ class View
 
     void setParent(View* parent, void* parentUserdata = nullptr);
     View* getParent();
+    std::vector<View*>& getChildren()
+    { return children; }
     bool hasParent();
 
     void* getParentUserData();
@@ -357,18 +360,6 @@ class View
     {
         return nullptr;
     }
-
-    /**
-     * Return the child at xpos and ypos (in screen coordinated), or null if
-     * there is none at the position.
-     */
-    virtual View* getChildViewAtTouch(float xpos, float ypos);
-
-    /**
-     * Self explanatory.
-     */
-    virtual bool isFocusable() const
-    { return false; }
 
     /**
      * Self explanatory.
--- a/library/lib/application.cpp
+++ b/library/lib/application.cpp
@@ -132,55 +132,58 @@ static void windowKeyCallback(GLFWwindow* window, int key, int scancode, int act
     }
 }
 
-struct CursorPos {
-    double x{ NAN };
-    double y{ NAN };
-};
-
-CursorPos last_cursor_pos;
-CursorPos current_cursor_pos;
-CursorPos delta_cursor_pos;
-View* current_draggable{ nullptr };
-
 static void cursorPosCallback(GLFWwindow* window, double xpos, double ypos){
+    auto& last_cursor_pos{ Application::getLastPointerPos() };
+    auto& current_cursor_pos{ Application::getCurrPointerPos() };
+    auto& delta_cursor_pos{ Application::getDeltaPointerPos() };
+
     last_cursor_pos = current_cursor_pos;
-    glfwGetCursorPos(window, &current_cursor_pos.x, &current_cursor_pos.y);
+    double x;
+    double y;
+    glfwGetCursorPos(window, &x, &y);
+    current_cursor_pos.x = x;
+    current_cursor_pos.y = y;
     delta_cursor_pos.x = current_cursor_pos.x - last_cursor_pos.x;
     delta_cursor_pos.y = current_cursor_pos.y - last_cursor_pos.y;
 
-    if(current_draggable && (delta_cursor_pos.x || delta_cursor_pos.y))
-        // delta_cursor_pos is multiplied by 5 (which is a random number that works ok.)
-        current_draggable->dragView(delta_cursor_pos.x * 5, -delta_cursor_pos.y * 5);
+    if(glfwGetMouseButton(window, GLFW_MOUSE_BUTTON_LEFT) == GLFW_PRESS
+    && (delta_cursor_pos.x || delta_cursor_pos.y)
+    ){
+        TouchEvent touch_event;
+        touch_event.type = TouchEventType::DRAG;
+        touch_event.pos = current_cursor_pos;
+        touch_event.delta = delta_cursor_pos;
+        // Set the touch event.
+        Application::getTouchEvent() = std::move(touch_event);
+    }
     //Logger::debug("Mouse pos: (x: %lf, y: %lf)", current_cursor_pos.x, current_cursor_pos.y);
     //Logger::debug("Mouse delta: (x: %lf, y: %lf)", delta_cursor_pos.x, delta_cursor_pos.y);
 }
 
-bool is_mouse_left_down{ false };
-CursorPos cursor_drag_start_pos;
-
 static void mouseBtnCallback(GLFWwindow* window, int button, int action, int mods){
+    auto& current_cursor_pos{ Application::getCurrPointerPos() };
+    auto& delta_cursor_pos{ Application::getDeltaPointerPos() };
+
     switch(button){
         case GLFW_MOUSE_BUTTON_LEFT:
             if(action == GLFW_PRESS){
-                is_mouse_left_down = true;
-                cursor_drag_start_pos = current_cursor_pos;
-                current_draggable = Application::getDraggable(cursor_drag_start_pos.x, cursor_drag_start_pos.y);
+                TouchEvent touch_event;
+                touch_event.type = TouchEventType::TOUCH;
+                touch_event.pos = current_cursor_pos;
+                touch_event.delta = delta_cursor_pos;
+                Application::getTouchEvent() = std::move(touch_event);
                 Logger::debug("Mouse Left down.");
             } else
             // is equivalent to "else if(action == GLFW_RELEASE" since action can
             // only be GLFW_PRESS or GLFW_RELEASE
             {
-                current_draggable = nullptr;
-                is_mouse_left_down = false;
-                const double xdrag{ current_cursor_pos.x - cursor_drag_start_pos.x };
-                const double ydrag{ current_cursor_pos.y - cursor_drag_start_pos.y };
-
-                if(!(xdrag || ydrag))
-                    // Try to focus on the position of the mouse up.
-                    Application::giveFocus(current_cursor_pos.x, current_cursor_pos.y);
+                TouchEvent touch_event;
+                touch_event.type = TouchEventType::RELEASE;
+                touch_event.pos = current_cursor_pos;
+                touch_event.delta = delta_cursor_pos;
+                Application::getTouchEvent() = std::move(touch_event);
 
                 Logger::debug("Mouse Left up.");
-                Logger::debug("Mouse Left drag delta (x: %lf, y: %lf)", xdrag, ydrag);
             }
         break;
     }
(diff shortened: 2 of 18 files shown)


library/lib/application.cpp
```diff
@@ -451,9 +511,18 @@ void Application::navigate(FocusDirection direction)
 {
     View* currentFocus = Application::currentFocus;
 
-    // Do nothing if there is no current focus or if it doesn't have a parent
+    // We should allow the user to regain focus to a default focus view,
+    // in the most recent view in the stack, in a case where the user is
+    // unable to navigate through normal means, because currentFocus is nullptr.
+    // It will happen with mouse and touch controls, it has happened to me.
+    if(!currentFocus){
+        Application::giveFocus(viewStack.back()->getDefaultFocus());
```
natinusala, line 519: That will cause issues for modal dialogs, with that the user will be able to focus what's under the dialog, that's not a good idea

library/lib/application.cpp
```diff
@@ -132,6 +132,64 @@ static void windowKeyCallback(GLFWwindow* window, int key, int scancode, int act
     }
 }
 
+struct CursorPos {
```
natinusala, line 135: Not sure if we want that dumped in the cpp like that :eyes:

natinusala:
Thanks for the PR!

I wonder if there is not a simpler and maybe more efficient way of doing that. Instead of iterating over all views when a glfw touch callback is executed, I thought about just storing the xpos and ypos of the latest touch event, and handling it in each view independently in the main loop (in the frame() method). If you think about it, touch events only concern the touched view, others don't care about what's happening there.

library/lib/application.cpp
jon-dez: I've tested it with the Dialog class and that doesn't seem to be a concern. I think the reason for it is due to the way modals are processed on the view stack.
```
// dialog.cpp
void Dialog::open()
{ // Line 58
    ...
    Application::pushView(this); // The modal dialog is pushed to the top of the stack. It is now the most recent view on the stack.
    ...
}
```
The modifications I did only touch the modal view because it is the most recent view on the stack, and it selects whatever button is the default focus within that view. With mouse and touch, the same thing happens. A touch or click queries the most recent view on the stack so the same principle applies.

library/lib/application.cpp
jon-dez: Yea I was unsure myself. I mean I see no negative removing it and just using the two doubles by themselves. I'll be honest and just say I was feeling lazy and wanted to save on typing.

It's your call, I can tidy things up by removing it or placing it somewhere else.

library/lib/application.cpp
natinusala: Did you try a dialog without any button though? Because that's a thing

library/lib/application.cpp
natinusala: I mean all this stuff here, all the variables, why are they not Application class members?

jon-dez:
> Thanks for the PR!
> 
> I wonder if there is not a simpler and maybe more efficient way of doing that. Instead of iterating over all views when a glfw touch callback is executed, I thought about just storing the xpos and ypos of the latest touch event, and handling it in each view independently in the main loop (in the frame() method). If you think about it, touch events only concern the touched view, others don't care about what's happening there.

Hmm... I see what you are saying. We could store the touch event and have it be processed in the Application::mainLoop() instead of whenever the callback is issued. So a revised processed will look something like this in the mainLoop?
```


bool Application::mainLoop(){
    // stuff before
    ...
    switch(touch_event.type){
        case TOUCH:
            Application::giveFocus(touch_event.x, touch_event.y);
        break;
        case DRAG:
            if(View* draggable{ Application::getDraggable(touch_event.x, touch_event.y) })
                draggable->dragView(touch_event.dx, touch_event.dy);
        break;
        ... // Others perhaps?
    }
    ...
    // stuff after
}
```

I'm STRONGLY against doing the touch processing in the frame() method of the views, since that just adds another layer of complexity to the frame() method and will be a pain to deal with in the future when we want to extend the capabilities of the touch system.
The method that is in place right now is already pretty simple and efficient because it is essentially a simple querying mechanism. We start off at the parent view, and ask a child one by one if the touch lies within its boundaries. If a child replies yes, then we check its children. Rinse and repeat. Also, the same mechanism works for both touch and drag events.
It's pretty minimal and efficient already if you ask me :smile:

jon-dez:
Also the whole "touch checking" (or boundary checking) algorithm isn't recursive. It is iterative as long as any getChildViewAtTouch(...) overrides do not itself call getChildViewAtTouch(...), which it shouldn't because it does not need to.

library/lib/application.cpp
jon-dez: No, but I will do that in one minute. Please hold.

jon-dez: Ok I just did. It was not possible to click outside the modal to make it go away.

But since the modal dialogs don't have any default focus I could not find a way to make the dialog disappear. I don't think it is possible with keyboard and mouse to escape a modal dialog if it does not have buttons.

library/lib/application.cpp
jon-dez: That is a good point, they should be Application class members.

On it captain!

jon-dez:
I made an edit to the pseudocode above in case you had the previous in mind. It occurred to me that we should query the draggable when a drag event is FIRST captured by glfw, if not then the previous code will try to query a draggable at a new position every frame, which may be a different view from when the drag started.

library/lib/application.cpp
natinusala: > But since the modal dialogs don't have any default focus I could not find a way to make the dialog disappear. I don't think it is possible with keyboard and mouse to escape a modal dialog if it does not have buttons.

This is actually intended, to create modal loading dialogs for instance.

library/include/borealis/view.hpp
```diff
@@ -351,6 +358,29 @@ class View
...
+    /**
+     * Return the child at xpos and ypos (in screen coordinated), or null if
+     * there is none at the position.
+     */
+    virtual View* getChildViewAtTouch(float xpos, float ypos);
+
+    /**
+     * Self explanatory.
+     */
+    virtual bool isFocusable() const
```
natinusala, line 370: I am not a fan of redefining `isFocusable` - we already have a definition of something focusable, it's "getDefaultFocus() returns this". I think this new method should be called "isTouchable" or "isFocusableByTouch" maybe?

library/lib/view.cpp
```diff
@@ -627,4 +627,13 @@ void View::invalidate(bool immediate)
         this->dirty = true;
 }
 
+View* View::getChildViewAtTouch(float xpos, float ypos){
+    Logger::debug("Touch position (x: %f, y: %f): %s does not provide a getChildViewAtTouch(...) override.", xpos, ypos, describe().c_str());
+    return nullptr;
+}
+
+void View::dragView(float dx, float dy){
+    Logger::debug("Drag performed on %s, but it does nothing.", describe().c_str());
```
natinusala, line 636: This will probably be a perf killer :eyes:

library/lib/view.cpp
```diff
@@ -627,4 +627,13 @@ void View::invalidate(bool immediate)
         this->dirty = true;
 }
 
+View* View::getChildViewAtTouch(float xpos, float ypos){
```
natinusala, line 630: This method could be made generic if we had a generic parent -> child relationship next to the child -> parent one. In that case we can just iterate over all children of the view tree and find the right one directly. We can then get rid of each reimplementation of getChidViewAtTouch.

Would that be inefficient?

library/lib/application.cpp
```diff
@@ -456,6 +588,15 @@ void Application::navigate(FocusDirection direction)
     if (!currentFocus || !currentFocus->hasParent())
         return;
 
+    // The item was "loosely" unfocused, i.e. we just wanted the highlight
+    // to disappear. Now we need it to reappear because the gamepad is
+    // navigating
+    // Almost like HOS, its not perfect.
```
natinusala, line 594: How is it "almost" like HOS? What's the difference?

library/lib/application.cpp
```diff
@@ -353,8 +427,64 @@ bool Application::mainLoop()
         Application::exit();
         return false;
     }
+    // Poll Touches from the switch's touch screen.
```
natinusala, line 430: I think glfw switch already binds touch to glfw mouse :eyes: Can you update your switch-glfw package and make sure it's the case?

We'll get rid of glfw for switch eventually so it's actually good that you made that (cc @WerWolv )

library/lib/application.cpp
```diff
@@ -132,6 +132,76 @@ static void windowKeyCallback(GLFWwindow* window, int key, int scancode, int act
...
+
+    if(glfwGetMouseButton(window, GLFW_MOUSE_BUTTON_LEFT) == GLFW_PRESS
+    && (Application::getDeltaPointerPos().x || Application::getDeltaPointerPos().y)
+    ){
+        TouchEvent touch_event;
+        touch_event.type = TouchEventType::DRAG;
+        touch_event.pos = Application::getCurrPointerPos();
+        touch_event.delta = Application::getDeltaPointerPos();
+        // Set the touch event.
+        Application::getTouchEvent() = std::move(touch_event);
```
natinusala, line 161: I'm not against std::move but uh can we maybe use more standard OOP stuff (setTouchEvent)? I don't like assigning something to a method return value directly like that

library/lib/application.cpp
jon-dez: Ok, that makes sense.

library/include/borealis/view.hpp
jon-dez: I like the "isFocusableByTouch()," I'll remember to do that when I get the chance.

library/lib/view.cpp
jon-dez: Yea, I chose to put that there for debugging purposes. I'll make sure to go over all the Logger::debug() calls I added when I finish with the draft.

library/lib/view.cpp
jon-dez: Sorry if I am misunderstanding you, but If I understood correctly you want there to be a function call that returns an iterable list of children that could be used to find the correct child?
In other words replace **getChildViewAtTouch** with **getChildren?**

library/lib/application.cpp
jon-dez: Just the way it currently behaves.
When navigating by gamepad in HOS and then switching over to navigating by touch-screen causes the "focus" to disappear. Then when switch back to gamepad navigation by pressing "OK" (The 'A' button) causes the focus to reappear, well pressing any button makes that happen.
However in this draft, when switching back to gamepad navigation by pressing "OK" causes the onclick event to fire. Thus, the behavior is "almost" like HOS.

library/lib/application.cpp
jon-dez: Yea I was initially going down that route and using mouse implementation in glfw to make the touch-screen work, but there were some issues I came across that made me just go ahead and use the hid touch screen functions provided by libnx.

But what's the reasoning behind getting rid of glfw? Was it meant to be temporary?

library/lib/application.cpp
jon-dez: Ok I'll do that.

library/include/borealis/view.hpp
jon-dez: Actually no, scratch that. I think isTouchable describes its purpose better.

jon-dez: Quick question:
Should every focusable be a touchable? If so then I think it'd be best to just scrap the isFocusable/isFocusableByTouch idea that I had and just check if getDefaultFocus is equivalent to the object that is being check.

library/lib/application.cpp
WerWolv: The issue with glfw on the switch is that it's massive. Switching to deko means cutting down the executable size from like 7MB to about 500kB. This is great for homebrew but even better for when people eventually try to reimplement for example applets. I did some tests with borealis replacing qlaunch and it wouldn't even load the example without at least 150MB of heap allocated. This will also be resolved with deko.

library/lib/application.cpp
natinusala: Okay I see. I think the way it's implemented in the draft is actually bad design - people won't expect it to behave that way and it will cause troubles. It's the kind of little things that make you hate a piece of software as a user 😬

library/lib/view.cpp
natinusala: Remove getChildViewAtTouch in every View implementation and replace it with a loop that traverses the view tree, starting with the root flowing down to the children.

But to do that you need a generic parent -> child relation between views. Right now we only have parent -> child.

library/include/borealis/view.hpp
natinusala: Every focusable is a touchable but not every touchable is focusable (bottom-right hints for instance are touchable and not focusable)

WerWolv: Lists itself (not the items in it) are not focusable but touchable

library/lib/view.cpp
jon-dez: commit fccfeda is my take on this.

jon-dez: There is an Issue with TabFrame though, but I'm going to push another commit to fix that.

library/lib/application.cpp
jon-dez: I think the best way to approach this is to implement some type of "Navigation Controller," where navigation works a certain way based on the input method (gamepad/keyboard or touchscreen.)

library/lib/view.cpp
natinusala: I like the automated children detection in `setParent`. However the same view can be added twice - setParent should be idempotent. Could you use a data structure that doesn't allow duplicates ?

Also, the vector is never cleaned when the children are deleted so it looks like a hazard to me

jon-dez: When you say "cleaned" what do you mean?

natinusala: As in "deleted children are not removed from the vector"

jon-dez: Hmm, I see. I guess the problem is that there really isn't a system put in place that manages the relationship between child and parent views in a generalized way, and because of that it's really difficult to incorporate some sort of touch support without having to insert some specialization into many view classes so that it plays nicely with touches. What do you think?

library/lib/application.cpp
natinusala: Can't you simply implement the missing behavior? With a touch mode boolean for instance. When pressing a button if in touch mode, don't call the click event but simply disable touch mode and show the focus back ?

natinusala:
Hi, sorry for the delay!

I don't accept contributions for master anymore, so I will be closing this issue.

I am rewriting the library starting (almost) from scratch, and #77 already aims to provide touch support for that version of the library. Feel free to chime in and contribute to the discussion and/or code there if you want!

PR #40, opened 2020-07-06
TabFrame and Dialog modifications.

jon-dez:
TabFrame: rightPane now manages responsibilty for displaying the sidebar item's view.
Dialog: set horizontalButtonsLayout's parent.

This is somewhat related to the Touch pr draft such that these modifications fix some problems with "touching" views.
This is separate from the Touch pr draft, since these changes do not depend on the Touch pr, and  could be merged without it.

### DIFF
--- a/library/lib/tab_frame.cpp
+++ b/library/lib/tab_frame.cpp
@@ -27,15 +27,52 @@
 namespace brls
 {
 
+void ViewContainer::setView(View* view){
+    //children.clear(); // Clear the only possible child.
+    if(associated_view){
+        associated_view->willDisappear(true);
+    }
+    if(view) {
+        view->setParent(this);
+        view->willAppear(true);
+    }
+    associated_view = view;
+    invalidate();
+}
+
+void ViewContainer::draw(NVGcontext* vg, int x, int y, unsigned width, unsigned height, Style* style, FrameContext* ctx) {
+    if(associated_view)
+        associated_view->frame(ctx);
+}
+
+
+void ViewContainer::layout(NVGcontext* vg, Style* style, FontStash* stash)
+{
+    if(associated_view) {
+        associated_view->setBoundaries(x, y, width, height);
+        associated_view->invalidate();
+    }
+}
+
+View* ViewContainer::getDefaultFocus() {
+    if(associated_view)
+        return associated_view->getDefaultFocus();
+    return nullptr;
+}
+
 TabFrame::TabFrame()
     : AppletFrame(false, true)
 {
     //Create sidebar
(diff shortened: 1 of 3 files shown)


library/lib/tab_frame.cpp
```diff
@@ -27,15 +27,52 @@
 namespace brls
 {
 
+void ViewContainer::setView(View* view){
+    //children.clear(); // Clear the only possible child.
```
jon-dez, line 31: This should be uncommented when merging with the Touch support pr branch.

natinusala:
Thanks for the PR!

I'm sorry to say that now but we already intended to rewrite TabFrame using the LayerView @WerWolv made. This achieves roughly the same thing (LayerView would replace the container you wrote).

Do you think you could adapt your PR to use LayerView instead?

jon-dez:
Adapting the PR to use LayerView seems more complex.
My reasoning is the way I see it is that the Sidebar is already some sort of LayerView, since it contains SidebarItems's with associated views. I see the SidebarItems's as layers that ask the "rightPane" (ViewContainer) to bring it into view when they are clicked.

natinusala:
How is the Sidebar a LayerView ? I would just replace the right pane with LayerView and have the sidebar call changeLayer when needed ?

jon-dez:
I guess what I meant to say was that SideBarItems have an associated view with them, and we can use their click callbacks to change "rightPane" like so:

```
...
// Using ViewContainer
// 'item' is a SideBarItem
item->getFocusEvent()->subscribe([this](View* view) {
    if (SidebarItem* item = dynamic_cast<SidebarItem*>(view))
       this->rightPane->setView(item->getAssociatedView());
});
...
```
But using LayerView:
```
...
// Using TabFrame
// 'item' is a SideBarItem
int next_item_idx = this->rightPane->getLayerCount(); // Note: This member function does not exist
this->rightPane->addLayer(item->getAssociatedView());
item->getFocusEvent()->subscribe([this, next_item_idx](View* view) {
    if (SidebarItem* item = dynamic_cast<SidebarItem*>(view))
       this->rightPane->changeLayer(next_item_idx, true);
});
...
```
As you can see, there are less steps involved in the first example. Also, it eliminates having to keep track of the index of the sidebar item's associated view in the view layer and avoids having to keep a record of the views in a LayerView.

natinusala:
Why don't you add the associated view id in the SidebarItem ? When adding a tab, add the view to the LayerView and store the id in the SidebarItem

natinusala:
Closed as this PR is reimplemented or made irrelevant by the changes on the `yoga` branch.

PR #43, opened 2020-07-07
Added another wrapper for List...

ckurtz22:
Added wrapper for getViewsCount so you can iterate over children externally.
Moved getSidebar from protected to public. I'm not sure if it was intended to only make inherited classes of ThumbnailFrame?


natinusala:
Thanks for the PR! I don't remember about Sidebar but I don't mind exposing more stuff

PR #47, opened 2020-07-18
Add description parameter to SelectListItem constructor

ConsoleLogLuke:
All of the list item types have a `description` parameter in their constructors except for `SelectListItem`, so I added it to `SelectListItem` to ensure consistency and obviously make it easier to add a description to a `SelectListItem`.

### DIFF
--- a/library/include/borealis/list.hpp
+++ b/library/include/borealis/list.hpp
@@ -120,7 +120,7 @@ typedef Event<int> ValueSelectedEvent;
 class SelectListItem : public ListItem
 {
   public:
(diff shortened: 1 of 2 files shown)


natinusala:
Thanks for the PR!

How do they look in the app? It shouldn't change anything but I just want to make sure it doesn't break something visually

ConsoleLogLuke:
I checked and everything looks fine, it behaves just like descriptions behave with other list item types!

natinusala:
Okay great!

PR #48, opened 2020-07-19
Fix visual overflowing of sidebar items

Jpe230:
Adding more than 7 items to the sidebar results in the remaining items to be visually overflowed past the viewbox

This PR serves two functions:

1. Fix the overflowing of items by using ListView as the base class for the Sidebar View
2. Incorporate LayerView in TabFrame to take advantages of the animations


library/include/borealis/tab_frame.hpp
```diff
@@ -23,10 +23,13 @@
 #include <borealis/sidebar.hpp>
 #include <string>
 #include <vector>
+#include <map>
 
 namespace brls
 {
 
+//fwd declaration
+class LayerView;
```
natinusala, line 32: Why don't you just import layer_view.hpp ? 🤔

library/include/borealis/tab_frame.hpp
```diff
@@ -51,9 +54,10 @@ class TabFrame : public AppletFrame
   private:
     Sidebar* sidebar;
     BoxLayout* layout;
-    View* rightPane = nullptr;
+    LayerView* rightPane = nullptr;
+    std::map<u_int64_t, int> panelMap;
 
-    void switchToView(View* view);
+    void switchToView(u_int64_t viewIndex);
```
natinusala, line 60: We usually use `u64` for integer types

library/include/borealis/tab_frame.hpp
```diff
@@ -51,9 +54,10 @@ class TabFrame : public AppletFrame
   private:
     Sidebar* sidebar;
     BoxLayout* layout;
-    View* rightPane = nullptr;
+    LayerView* rightPane = nullptr;
+    std::map<u_int64_t, int> panelMap;
```
natinusala, line 58: Why the need for such a map? Can't you just use the int id straight from LayerView?

natinusala:
Thanks for the PR!

I didn't read all the way through yet because there are things I want to clarify first, that may require some refactoring on your end. Check out my comments on the code 😄 

Also there is one thing I don't like, it's that it creates a coupling between Sidebar and TabFrame. What would you think of turning that associatedViewId (in SidebarItem) into a generic "user data" pointer? That way, Sidebar can still be used with something else than TabFrame.

Jpe230:
Hope you like it 👍

library/lib/sidebar.cpp
```diff
@@ -37,40 +37,34 @@ Sidebar::Sidebar()
...
-SidebarItem* Sidebar::addItem(std::string label, View* view)
+SidebarItem* Sidebar::addItem(std::string label, void* userdata)
 {
     SidebarItem* item = new SidebarItem(label, this);
-    item->setAssociatedView(view);
+    item->setAssociatedViewUserData(userdata);
 
-    if (this->isEmpty())
+    /*focus first item as soon as we add it*/
+    if (currentActive == nullptr)
```
natinusala, line 63: this->

natinusala: actually if (!currentActive)

library/lib/sidebar.cpp
```diff
@@ -37,40 +37,34 @@ Sidebar::Sidebar()
...
         setActive(item);
 
     this->addView(item);
+
```
natinusala, line 67: ❌

library/include/borealis/sidebar.hpp
```diff
@@ -45,8 +45,9 @@ class SidebarItem : public View
     std::string label;
     bool active = false;
 
-    Sidebar* sidebar     = nullptr;
-    View* associatedView = nullptr;
+    Sidebar* sidebar              = nullptr;
```
natinusala, line 48: I wonder if we can go even further. Remove SidebarItem entirely, add a single void* userdata in ListItem.

Use ListItem instead of SidebarItem in Sidebars, in the userdata put the three things we need: sidebar, associatedView, associatedViewUserData

Backport all features of ListItem in SidebarItem (if there is any?)

What do you think?

library/lib/tab_frame.cpp
```diff
@@ -52,36 +57,49 @@ bool TabFrame::onCancel()
     return AppletFrame::onCancel();
 }
 
-void TabFrame::switchToView(View* view)
+void TabFrame::switchToView(u_int64_t viewIndex)
```
natinusala, line 60: u64

library/include/borealis/sidebar.hpp
```diff
@@ -45,8 +45,9 @@ class SidebarItem : public View
...
+    View* associatedView          = nullptr;
+    void* associatedViewUserData  = nullptr;
```
natinusala, line 50: free it in dtor

library/lib/sidebar.cpp
```diff
@@ -37,40 +37,34 @@ Sidebar::Sidebar()
...
 {
-    // Sanity check
-    if (this->lastFocus >= this->children.size())
-        this->lastFocus = 0;
-
     View* toFocus{ nullptr };
     // Try to focus last focused one
-    if(this->children.size() != 0)
-        toFocus = this->children[this->lastFocus]->view->getDefaultFocus();
+    if(this->currentActive != nullptr)
```
natinusala, line 42: if (this->currentActive)

library/lib/tab_frame.cpp
```diff
@@ -52,36 +57,49 @@ bool TabFrame::onCancel()
...
-    this->rightPane = view;
+    
     if (this->rightPane != nullptr)
         this->layout->addView(this->rightPane, true, true); // addView() calls willAppear()
 }
 
 void TabFrame::addTab(std::string label, View* view)
 {
-    SidebarItem* item = this->sidebar->addItem(label, view);
+
```
natinusala, line 81: ❌

library/include/borealis/sidebar.hpp
```diff
@@ -68,19 +69,22 @@ class SidebarItem : public View
...
 
+    void   setAssociatedViewUserData(void* userdata);
+    void*  getAssociatedViewUserData();
+
     ~SidebarItem();
 };
 
 // TODO: Add a style with icons, make it collapsible?
-class Sidebar : public BoxLayout
+class Sidebar : public List
```
natinusala, line 79: Should Sidebar even exist? Isn't it only a List with different styling and the active thing? Could it not be all merged into List?

Jpe230: Asked the same question myself, how should I tackle this?

1) Remove class entirely and apply margins via a new method inside List
2) Remove class entirely and apply margins via an overloaded constructor inside List
3) Remove class entirely and apply margins via TabFrame
4) Keep Sidebar to override constructor from List to apply margins

natinusala: I'd say a mix of 1 and 4

- Empty Sidebar class entirely and apply margins via a new method inside List
- Call the new margins method in the Sidebar constructor

Jpe230: I think I got it, kinda tricky had to backport several things, but it works amazing

natinusala:
Can you resolve the conflicts please?

natinusala:
Sorry for the delay! I will be closing this PR as I don't accept contributions for master anymore, and the issue has been independently fixed in the rework branch.

Issue #27, opened 2020-06-19
Focus changes can produce wrong view state


natinusala:
Pressing UP and DOWN at the same time gives this:

![image](https://user-images.githubusercontent.com/1159464/85114992-42097a80-b21b-11ea-8827-573771e85f65.png)

Both views are in the focused state, which is wrong.

natinusala:
Fixed by #54

natinusala:
#54 must be ported to yoga (keep authorship)

natinusala:
Actually the "input block tokens" can / should be used instead of the changes in #54

PR #52, opened 2020-08-03
Use sticks to navigate

Chrscool8:
[Demo Vid](https://youtu.be/lTX_jIK_vpI)

Pretty straight-forward, quick and easy chunk of code that sets the d-pad nav values based on the stick states. Works perfect and feels great except for one thing.

The main problem is that [other, more wide-spread problem](https://github.com/natinusala/borealis/issues/27#issue-641814645), where if you push two directions at the same time, multiple things can accept the focus. This is a lot easier to do on the same frame with a diagonally pressed stick than pushing two distinct d-pad buttons at the same time. I can add a band-aid where it'll only accept right/left stick inputs when not pushing up/down, but like I said, that's just a fix for this one case.

### DIFF
--- a/library/lib/application.cpp
+++ b/library/lib/application.cpp
@@ -381,7 +381,7 @@ bool Application::mainLoop()
     const float* axes = glfwGetJoystickAxes(GLFW_JOYSTICK_1, &count);
 
     //Make sure we got all our expected sticks before reading from them
-    if (count == 4)
+    if (count >= 4)
     {
         // Left Stick Y
         if (axes[1] < -.5)
--- a/library/lib/application.cpp
+++ b/library/lib/application.cpp
@@ -375,7 +375,38 @@ bool Application::mainLoop()
     }
 
     // Trigger gamepad events
-    // TODO: Translate axis events to dpad events here
+
+    // Translate axis events to dpad events
+    int count;
+    const float* axes = glfwGetJoystickAxes(GLFW_JOYSTICK_1, &count);
+
+    //Make sure we got all our expected sticks before reading from them
+    if (count == 4)
+    {
+        // Left Stick X
+        if (axes[0] < -.5)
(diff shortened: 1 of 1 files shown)


Chrscool8:
Band-aid commit above prioritizes up and down over left and right. It feels almost exactly the same but will never do the split-up as long as you're only using one type of control at once.

natinusala:
Hey, I like that thanks!

Did you try on PC? How does it behave when using directional keys?

Also I think there should be a focus lock that prevents a focus change when another is going, but at the focus system level, not at the input level. There can be as many inputs as we want at the same time in one frame, the system should be able to handle that.

Chrscool8:
Thanks!

Not sure. I've never actually been able to compile for Windows because I can't figure out how to get my msys2 to find and install the mingw packages required. :\

Agreed on that last part!

natinusala:
If you use the mingw msys2 everything will work out of the box

Chrscool8:
Nice, got most of that going, just trying to get it to find glfw3 now.

Chrscool8:
Got it going on PC! Works great with an Xbox controller now, too.

natinusala:
Nice, did you tackle the multiple focus issue?

Chrscool8:
That's still working with the band-aid. Gonna look into that deeper soon!

Chrscool8:
Card link just for handy reference:
https://github.com/natinusala/borealis/projects/1#card-25958217

natinusala:
Can you rebase on master once the other PR has been merged? Thanks :D

Chrscool8:
Can do!

natinusala:
I will close this PR as I don't accept contributions for master anymore.

However I will port your changes to the `yoga` branch, while keeping authorship. Unless you want to do it yourself?

Chrscool8:
You're welcome to pull 'em over yourself if it isn't too much trouble! Hoping to help with the new version soon when I get back into dev stuff. Thanks!

PR #53, opened 2020-08-14
Add draw clipping

ITotalJustice:
Only draw what's visible on screen. It can be toggled by the user for if they use a box container that needs to be drawn which has children which would be visible on screen. An example of this is the `ListContentView` class (I have updated the class).

I'll mention how the clipping works in the documentation later for the users.

### DIFF
--- a/library/include/borealis/view.hpp
+++ b/library/include/borealis/view.hpp
@@ -101,7 +101,7 @@ class View
 
     bool hidden = false;
 
-    bool clipping = true;
+    bool clipped = true;
 
     std::vector<Action> actions;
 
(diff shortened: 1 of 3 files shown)


library/include/borealis/view.hpp
```diff
@@ -101,6 +101,8 @@ class View
 
     bool hidden = false;
 
+    bool clipping = true;
```
natinusala, line 104: would it not be "clipped" instead of "clipping"? or "clippingEnabled"?

natinusala:
Thanks for the PR!

You should take width and height into account too, that way you don't have to setClipping(true) the Lists

library/include/borealis/view.hpp
ITotalJustice: good point!

natinusala:
Closed as this PR is reimplemented or made irrelevant by the changes on the `yoga` branch. Clipping is automated by the Box class.

PR #54, opened 2020-08-15
Focus splitting fixes

Chrscool8:
Got it! Turned out to be surprisingly simple once I learned a bit how it worked more deeply.

This fixes https://github.com/natinusala/borealis/issues/27 by using a simple bool lock.

The lock is activated when/if the focus is successfully changed and deactivated at the end of the main loop each step. This way, only one view change can happen in one frame. I can't find any weirdness or problems with this solution in extensive testing.

This also gracefully fixes the problem with the sticks in https://github.com/natinusala/borealis/pull/52 and removes the need for its 'band-aid' solution.

### DIFF
--- a/library/lib/application.cpp
+++ b/library/lib/application.cpp
@@ -442,6 +442,8 @@ bool Application::mainLoop()
         }
     }
 
+    focusLocked = false;
+
     return true;
 }
 
@@ -452,6 +454,14 @@ void Application::quit()
 
 void Application::navigate(FocusDirection direction)
 {
+    printf("Navigate: %i\n", direction);
+
+    if (focusLocked)
+    {
+        printf("Locked\n");
+        return;
+    }
+
     View* currentFocus = Application::currentFocus;
 
     // Do nothing if there is no current focus or if it doesn't have a parent
@@ -479,6 +489,7 @@ void Application::navigate(FocusDirection direction)
     }
 
     // Otherwise give it focus
+    focusLocked = true;
     Application::giveFocus(nextFocus);
 }
 
--- a/library/lib/application.cpp
+++ b/library/lib/application.cpp
@@ -454,13 +454,8 @@ void Application::quit()
 
 void Application::navigate(FocusDirection direction)
 {
-    printf("Navigate: %i\n", direction);
-
     if (focusLocked)
-    {
-        printf("Locked\n");
         return;
-    }
 
     View* currentFocus = Application::currentFocus;
 
--- a/library/lib/application.cpp
+++ b/library/lib/application.cpp
@@ -442,7 +442,7 @@ bool Application::mainLoop()
         }
     }
 
-    focusLocked = false;
+    Application::focusLocked = false;
 
     return true;
 }
@@ -454,7 +454,7 @@ void Application::quit()
 
 void Application::navigate(FocusDirection direction)
 {
-    if (focusLocked)
+    if (Application::focusLocked)
         return;
 
     View* currentFocus = Application::currentFocus;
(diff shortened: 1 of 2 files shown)


library/lib/application.cpp
```diff
@@ -442,6 +442,8 @@ bool Application::mainLoop()
         }
     }
 
+    focusLocked = false;
```
natinusala, line 445: `Application::`

library/lib/application.cpp
```diff
@@ -479,6 +484,7 @@ void Application::navigate(FocusDirection direction)
     }
 
     // Otherwise give it focus
+    focusLocked = true;
```
natinusala, line 487: `Application::`

natinusala:
Thanks for the PR! I left some nitpicky comments :D

I'll have to test this extensively to make sure it doesn't break some sidebar behavior (like press left/right to move, a/b to enter and exit etc...).

natinusala:
Also I'll update the wiki to state that only one focus change can happen per frame.

A more elegant solution would probably to defer all calls to navigate() and only actually run the latest of the frame (at the end of the frame). This might break my fancy pants WIP scrolling changes though.

Chrscool8:
> Thanks for the PR! I left some nitpicky comments :D

Happy to have the tips! Those are done and done!
 
> I'll have to test this extensively to make sure it doesn't break some sidebar behavior (like press left/right to move, a/b to enter and exit etc...).

Have at it! I've tried it on pretty much every view and interaction type I know of and haven't seen anything iffy yet.

> Also I'll update the wiki to state that only one focus change can happen per frame.
> 
> A more elegant solution would probably be to defer all calls to navigate() and only actually run the latest of the frame (at the end of the frame). This might break my fancy pants WIP scrolling changes though.

For sure. I was toying around with that sorta idea and thinking about what might be good built-in key priorities but figured this went great as a temp solution at the very least!

natinusala:
Do you have Discord? If so, can you join the RetroNX server and DM me?

Chrscool8:
Anything else this and #54 needed before being added to the main code? (I'm finally back on a PC to keep working on stuff!)

natinusala:
No I just need to take care of it 🙈

natinusala:
@Chrscool8 Wow so it's really been three months since I said I would review it huh

The thing is, I am currently focusing my work on the yoga rework, and I'm not sure if I take contributions for the master branch?

So much stuff has changed and the two versions are basically incompatible now.

Are you working on something in particular with borealis now? Would you consider switching to the new version?

Chrscool8:
I... had no idea that was a thing. 😅

Just thought it would be nice if this and the other quick change would be merged so all the existing borealis apps out there could instantly update to use control sticks. 

I just moments ago finished a humongous rewrite of my borealis-incorporating app, if that's what you mean, which does scare me to learn that the new version is going to be so different... but I can check it out, for sure.

natinusala:
👀 Well I hope you're not too scared of rewriting it _again_ because I'm eventually going to drop support for the current version and go all in for the rework.

What part takes the UI in your code?

Chrscool8:
[Most](https://i.imgur.com/ikEvAJb.jpg) of [it](https://i.imgur.com/mOEk5gc.jpg). 🙃

I'll dive into the new version soon. I guess I'll just polish of what I can do with this one so far.

natinusala:
Okay well I see that you are using component that are not yet reimplemented in the new version so I would advise to keep using the old one for now :D

Chrscool8:
Good deal! Good luck with the new project! I'll keep an eye on it!

natinusala:
Unless... you want to help me out reimplementing everything? 😏

Chrscool8:
We'll be in touch lol

natinusala:
FYI I will close this PR and apply this commit on `yoga` instead (keeping the authorship).

PR #55, opened 2020-08-22
Allow ListItem to use View types as thumbnails.

meganukebmp:
This change allows the use of View items as thumbnails while still preserving all previous functionality, as to not break anything that might depend on this.

There is some ugliness _in my opinion_ to how I did this, mainly the dynamic casts and some repetition, but I feel as this is the cleanest way to do things. Rather have a few lines of clear repetition than a convoluted interlinking mess.

Attached is a testing application [main.txt](https://github.com/natinusala/borealis/files/5112915/main.txt)

### DIFF
--- a/library/include/borealis/list.hpp
+++ b/library/include/borealis/list.hpp
@@ -50,7 +50,9 @@ class ListItem : public View
     bool drawTopSeparator = true;
 
     Label* descriptionView = nullptr;
-    Image* thumbnailView   = nullptr;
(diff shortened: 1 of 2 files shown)


natinusala:
Thanks for the PR! Very sorry for the delay.

Unfortunately I am rewriting the library and I don't accept contributions for the master branch anymore, which will soon become legacy.

The rework does not have ListItem reimplemented yet so I cannot port your changes either... I will have another look when this is done.

Issue #56, opened 2020-09-10
get selected value from SelectListItem and block exit button


D3fau4:
Hello!
is there any way to get a value from any listitem?
if i cant get the value could you make public `unsigned selectedValue;`?
also i need block "+" because when i start the system update i can exit from the app. 
(I don't know if borealis already has that options.)
thanks!

natinusala:
Hello,
What kind of ListItem ?

D3fau4:
this type https://github.com/natinusala/borealis/blob/70bd48cdd1706c5dc6471246dac15fa026cc1f72/example/main.cpp#L85

natinusala:
Yeah it needs to be updated to add an `int getSelectedValue`.

As for disabling + to quit, you can do it with `brls::Application::setGlobalQuit(true)`

PR #57, opened 2020-09-10
Add getSelectedValue to SelectListItem

D3fau4:
#56

### DIFF
--- a/library/include/borealis/list.hpp
+++ b/library/include/borealis/list.hpp
@@ -123,6 +123,7 @@ class SelectListItem : public ListItem
     SelectListItem(std::string label, std::vector<std::string> values, unsigned selectedValue = 0, std::string description = "");
 
     void setSelectedValue(unsigned value);
+    int getSelectedValue();
 
     ValueSelectedEvent* getValueSelectedEvent();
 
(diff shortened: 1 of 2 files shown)


library/include/borealis/list.hpp
```diff
@@ -123,6 +123,7 @@ class SelectListItem : public ListItem
     SelectListItem(std::string label, std::vector<std::string> values, unsigned selectedValue = 0, std::string description = "");
 
     void setSelectedValue(unsigned value);
+    int getSelectedValue();
```
natinusala, line 126: Can you replace that by `unsigned` please?

natinusala (changes requested):
Thanks for the PR!

library/include/borealis/list.hpp
D3fau4: done

natinusala:
I see that selectedValue isn't initialized in the class, that's my fault, but can you add `= 0;` to the PR while you are here?

D3fau4:
Okay, I think it's done.

natinusala:
Great thank you!

natinusala (approved):

Issue #72, opened 2020-11-02
Allow users to manage swkbdConfigSetKeySetDisableBitmask 


H0neyBadger:
Hello, 
I think the borealis embedded swkbd is pretty cool.
unfortunately I'm using '/' for my project and this option is locked by default in borealis.
https://github.com/natinusala/borealis/blob/d5ba2eabc0b61a85075a37c21469e3cde5c6bae6/library/lib/swkbd.cpp#L65

I would like to propose a PR, but since swkbd flags are only known by the __SWITCH__ version,
I do not know how to properly implement this without altering the linux compatibility.
do you have any idea on how to proceed ?
thank you a lot

natinusala:
Hey there !

The answer would be to abstract out the swkbd capabilities in a new enum / bitmask, directly in borealis. Then, add a parameter for that in openForText and its friends. Finally, add a way to customize the capabilities on the different ListItems.

H0neyBadger:
OK I will make a PR for this.
Thank you for your guidance.

H0neyBadger:
thank you a lot

PR #73, opened 2020-11-02
Add enum kbdDisableBitmask to manage keyboard imput (#72)

H0neyBadger:
I'm keeping the PR in draft state to test the code correctly.
let me know if I have to change anything.
Thank you a lot

### DIFF
--- a/library/include/borealis/list.hpp
+++ b/library/include/borealis/list.hpp
@@ -24,6 +24,7 @@
 #include <borealis/label.hpp>
 #include <borealis/rectangle.hpp>
 #include <borealis/scroll_view.hpp>
+#include <borealis/swkbd.hpp>
 #include <string>
 
 namespace brls
@@ -162,9 +163,10 @@ class InputListItem : public ListItem
   protected:
     std::string helpText;
     int maxInputLength;
+    int kbdDisableBitmask;
 
   public:
-    InputListItem(std::string label, std::string initialValue, std::string helpText, std::string description = "", int maxInputLength = 32);
+    InputListItem(std::string label, std::string initialValue, std::string helpText, std::string description = "", int maxInputLength = 32, int kbdDisableBitmask = KeyboardKeyDisableBitmask::KEYBOARD_DISABLE_NONE);
 
     virtual bool onClick() override;
 };
@@ -174,7 +176,7 @@ class InputListItem : public ListItem
 class IntegerInputListItem : public InputListItem
 {
   public:
-    IntegerInputListItem(std::string label, int initialValue, std::string helpText, std::string description = "", int maxInputLength = 32);
+    IntegerInputListItem(std::string label, int initialValue, std::string helpText, std::string description = "", int maxInputLength = 32, int kbdDisableBitmask = KeyboardKeyDisableBitmask::KEYBOARD_DISABLE_NONE);
 
     virtual bool onClick() override;
 };
--- a/library/include/borealis/swkbd.hpp
+++ b/library/include/borealis/swkbd.hpp
@@ -25,11 +25,24 @@
 namespace brls
 {
 
+enum KeyboardKeyDisableBitmask
+{
+    KEYBOARD_DISABLE_NONE         = 0,
+    KEYBOARD_DISABLE_SPACE        = 1,
+    KEYBOARD_DISABLE_AT           = 1 << 1,
+    KEYBOARD_DISABLE_PERCENT      = 1 << 2,
+    KEYBOARD_DISABLE_FORWSLASH    = 1 << 3,
+    KEYBOARD_DISABLE_BACKSLASH    = 1 << 4,
+    KEYBOARD_DISABLE_NUMBERS      = 1 << 5,
+    KEYBOARD_DISABLE_DOWNLOADCODE = 1 << 6,
+    KEYBOARD_DISABLE_USERNAME     = 1 << 7,
+};
+
 class Swkbd
 {
   public:
-    static bool openForText(std::function<void(std::string)> f, std::string headerText = "", std::string subText = "", int maxStringLength = 32, std::string initialText = "");
-    static bool openForNumber(std::function<void(int)> f, std::string headerText = "", std::string subText = "", int maxStringLength = 32, std::string initialText = "", std::string leftButton = "", std::string rightButton = "");
+    static bool openForText(std::function<void(std::string)> f, std::string headerText = "", std::string subText = "", int maxStringLength = 32, std::string initialText = "", int kbdDisableBitmask = KeyboardKeyDisableBitmask::KEYBOARD_DISABLE_NONE);
+    static bool openForNumber(std::function<void(int)> f, std::string headerText = "", std::string subText = "", int maxStringLength = 32, std::string initialText = "", std::string leftButton = "", std::string rightButton = "", int kbdDisableBitmask = KeyboardKeyDisableBitmask::KEYBOARD_DISABLE_NONE);
 };
 
 } // namespace brls
--- a/resources/i18n/fr/main.json
+++ b/resources/i18n/fr/main.json
@@ -60,6 +60,18 @@
         "item1": "Item 1",
         "item2": "Item 2",
         "item3": "Item 3"
+    },
+
+    "keyboard": {
+        "string": {
+            "title": "Ouvrir le clavier",
+            "default": "Texte par défaut",
+            "help": "Message d'aide"
+        },
+        "number": {
+            "title": "Ouvrir le clavier numérique",
+            "help": "Message d'aide"
+        }
     }
 }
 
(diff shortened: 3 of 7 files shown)


library/include/borealis/list.hpp
```diff
@@ -162,9 +163,10 @@ class InputListItem : public ListItem
   protected:
     std::string helpText;
     int maxInputLength;
-
+    enum keyboard_disable_bitmask kbdDisableBitmask;
```
natinusala, line 166: Does it make sense to allow changing the bitmask for _all_ list items, rather than only for free input ones?

library/include/borealis/swkbd.hpp
```diff
@@ -22,14 +22,48 @@
...
 #include <string>
 
+#ifdef __SWITCH__
+#include <switch.h>
+#endif
+
 namespace brls
 {
 
+enum keyboard_disable_bitmask
```
natinusala, line 32: The point of making an abstraction is that you don't need to define the values twice - I would have only used the values after the #else in the enum, and then make another function to translate from the abstracted enum to the swkbd one, only for Switch.

library/include/borealis/swkbd.hpp
```diff
@@ -22,14 +22,48 @@
...
+{
+#ifdef __SWITCH__
+    KEYBOARD_DISABLE_NONE         = 0,
+    KEYBOARD_DISABLE_SPACE        = SwkbdKeyDisableBitmask_Space,        ///< Disable space-bar.
```
natinusala, line 36: No need for that funky `////<` syntax here

library/include/borealis/swkbd.hpp
natinusala: Also the name doesn't match the convention: `KeyboardKeyDisableBitmask` would fit better

natinusala (changes requested):
Thanks for the PR!

The formatting looks funky at places but otherwise it does the job. I left some comments, I let you review them.

Once you are finished with the changes, before force pushing the branch, can you run `./scripts/format.sh --fix` please?

library/include/borealis/swkbd.hpp
H0neyBadger: yes sorry I missed this, 
I copy/pasted the text from the libnx https://github.com/switchbrew/libnx/blob/3d4f61735ad8515dfb3e9dc0c0c7568babbdec33/nx/include/switch/applets/swkbd.h#L35
it might be use by an automated tool.

library/include/borealis/list.hpp
H0neyBadger: I modified the InputListItem. did I missed something ?
```
// A list item which spawns the swkbd
// to input its value (string)
class InputListItem : public ListItem
{
  protected:
    std::string helpText;
    int maxInputLength;
    enum keyboard_disable_bitmask kbdDisableBitmask;

```

H0neyBadger:
Hello,
regarding the default value :
```
SwkbdKeyDisableBitmask_At | SwkbdKeyDisableBitmask_Percent | SwkbdKeyDisableBitmask_ForwardSlash | SwkbdKeyDisableBitmask_Backslash
```

Currently (in this PR), all filters are disabled. should I port this filter as default value for swkbd ?

natinusala:
@WerWolv do you know why you disabled all "special" chars in swkbd by default? should we keep it that way or enable everything  back?

natinusala:
This is fine otherwise !

Let's wait for the answer from Wer and I'll try it once it's finished (and formatted 👀 ).

H0neyBadger:
Hi @natinusala,
thank you a lot for this review, 
by fixing the format, you mean the multiline on function declaration ?
let me know if you have any suggestion. I will fix it asap
```cpp
class Swkbd
{
  public:
    static bool openForText(std::function<void(std::string)> f, std::string headerText = "", std::string subText = "", int maxStringLength = 32, std::string initialText = "");
    static bool openForNumber(std::function<void(int)> f, std::string headerText = "", std::string subText = "", int maxStringLength = 32, std::string initialText = "", std::string leftButton = "", std::string rightButton = "");
    static bool openForText(std::function<void(std::string)> f, std::string headerText = "",
        std::string subText = "", int maxStringLength = 32, std::string initialText = "",
        int kbdDisableBitmask = KeyboardKeyDisableBitmask::KEYBOARD_DISABLE_NONE);
    static bool openForNumber(std::function<void(int)> f, std::string headerText = "",
        std::string subText = "", int maxStringLength = 32, std::string initialText = "",
        std::string leftButton = "", std::string rightButton = "",
        int kbdDisableBitmask = KeyboardKeyDisableBitmask::KEYBOARD_DISABLE_NONE);
};
```
I ran `./scripts/format.sh --fix` with clang-format 11 (default version on my distrib)

WerWolv:
@natinusala I think I just copied that function from my EdiZon codebase where I don't want these characters anywhere. You can absolutely change that behaviour

natinusala:
Okay then @H0neyBadger I'd like everything to be available by default (so nothing disabled)

As for the formatting, we use clang-format, there is a script available : `./scripts/format.sh --fix`

H0neyBadger:
yes, I already ran the fix `./scripts/format.sh --fix` but with clang-format 11 instead of clang-format-8.
I do not know if there is big difference. (clang-tools-extra-11.0.0-1.fc33.x86_64)
I can use a container with clang-format-8 to compare the result if needed

H0neyBadger:
I just tested with ubundu's clang-format-8 and there is nothing to fix anymore (the current proposed patches are not related to this PR).
As far as I can see, the `scripts/format.sh --fix` is ok.
But let me know if I have to change some formats/syntax/names.

natinusala:
So you're telling me that it's clang-format that splits the ListItem constructors in two lines? Same for openForNumber and friends

H0neyBadger:
Hi, I did the split first but clang-format do not complain about it. 
Should I put the ListItem construtors back in one line ?

H0neyBadger:
hi, all function declaration are back in one line (manually reverted). 
let me know if I have to reformat something. thank you a lot for your help.

H0neyBadger:
Hello @natinusala, 
sorry for asking. I know that's very difficult to find a spare moment for this.
Do you know if this change can be reviewed in a near future? 

I would like to improve our "chiaki" project & user inputs. 
But currently, I prefer to avoid workarounds in our code.
let me know If I have to implement this feature in our side.
thank you a lot

natinusala:
Hi, I am so sorry, I missed the latest changes you made, I always get an email and check as soon as I can but I never got the emails for that...

The code looks fine! Can I request one last thing (sorry), to add some example ListItems to the example app to showcase the different swkbd modes?

Thanks!

H0neyBadger:
Hello, thank you a lot for your reply, 
Ok no problem, I will add few examples ( like numbers and string ...)

btw some sxos users tend to report errors on `swkbdShow`. When they are using the regular homebrew channel the keyboard never show up. It affects all swkbdShow calls (not only borealis). the workaround is to run the homebrew channel in `app` mode, with full memory access.

H0neyBadger:
Hi, I just added an example.
Let me know if I have to change something.

resources/i18n/fr/main.json
```diff
@@ -60,6 +60,18 @@
         "item1": "Item 1",
         "item2": "Item 2",
         "item3": "Item 3"
+    },
+
+    "keyboard": {
+        "string": {
+            "title": "Ouvrir le Clavier",
+            "default": "Default text",
```
natinusala, line 68: `Texte par défaut`

resources/i18n/fr/main.json
```diff
@@ -60,6 +60,18 @@
         "item1": "Item 1",
         "item2": "Item 2",
         "item3": "Item 3"
+    },
+
+    "keyboard": {
+        "string": {
+            "title": "Ouvrir le Clavier",
```
natinusala, line 67: Lowercase C (Clavier -> clavier)

natinusala (changes requested):

H0neyBadger:
Nice catch. I've made the change requested

natinusala (approved):

natinusala:
Hey @H0neyBadger how are you?

I am currently in the process of changing the license of borealis from GPLv3 to Apache 2.0, and I need explicit permission from all contributors before continuing.

Do you allow me to change the license?

Thanks!

H0neyBadger:
Hi, I m fine and you ? ;-)
Yes, you have my explicit permission to change the license from GPLv3 to
Apache 2.0.

Is there any particular reason to change the license ? Does it has any
consequences for gpl3 projects that use borealis as gui ?

natinusala:
I'm great thank you😄 

The reason is that GPLv3 is objectively a bad choice for a library, it's too limiting. And people told me that when I released it.

With yoga I made the library portable by abstracting away everything platform specific, but I realized that it's useless if the code stays GPL because nobody would be able to port it to their own platform.

I would like the library to grow and for that I need to allow usage for everyone, even commercial projects.

As for your question, it should not change anything for projects using borealis as long as the license is compatible with Apache 2.0. If it's not then you are stuck with the latest GPL version.

The change is only for yoga anyway, so I think that you're safe for now with chiaki.

H0neyBadger:
Ok thank you a lot for this explanation.
I have to save some time to try the latest yoga version :) . Hope your
project will find the success it deserves.

I remember when I tried borealis for the first time,  the multiplatform was
a huge point in favour of borealis.

PR #49, opened 2020-07-19
Application: add Chinese shared font support

wwwwwwzx:
also works for traditional Chinese characters.
issue  #26

PR #76, opened 2021-01-20
Application: General i18n workaround for shared font loading

CaiMiao:
This is similar to #49 but I made a more graceful implement (at least imo), and I didn't aware of #49 when I was writing this lol

For memory consuming concern, we do not load full fallback chain when we are not in title takeover mode.

Confirmed working for my WIP tiny tool.

**I'm not really a programmer** so let me know if cleanup/optimization is needed.

---
Still I have a question, I can't really figure out how `nvgAddFallbackFontId` push the fallback order so I may have reversed my intend.

For example:
```
nvgAddFallbackFontId(Application::vg, Application::fontStash.regular, Application::fontStash.korean);
nvgAddFallbackFontId(Application::vg, Application::fontStash.regular, Application::fontStash.sharedSymbols);
```
does make the order `regular`->`korean`->`symbol` (appending) or `regular`->`symbol`->`korean` (prepending)?

### DIFF
--- a/library/include/borealis/i18n.hpp
+++ b/library/include/borealis/i18n.hpp
@@ -64,6 +64,14 @@ void loadTranslations();
  */
 std::string getCurrentLocale();
 
+#ifdef __SWITCH__
+/**
+ * Returns the current system locale id
+ * NOT the one that's currently used in the app!
+ */
+int swGetCurrentLocaleID();
+#endif
+
 inline namespace literals
 {
     /**
--- a/library/include/borealis/i18n.hpp
+++ b/library/include/borealis/i18n.hpp
@@ -64,13 +64,12 @@ void loadTranslations();
  */
 std::string getCurrentLocale();
 
-#ifdef __SWITCH__
 /**
  * Returns the current system locale id
+ * this id only make sense for libnx
  * NOT the one that's currently used in the app!
  */
-int swGetCurrentLocaleID();
-#endif
+int nxGetCurrentLocaleID();
 
 inline namespace literals
 {
--- a/library/lib/application.cpp
+++ b/library/lib/application.cpp
@@ -229,23 +229,126 @@ bool Application::init(std::string title, Style* style, LibraryViewsThemeVariant
 #ifdef __SWITCH__
     {
         PlFontData font;
+        int locale = i18n::swGetCurrentLocaleID();
 
         // Standard font
         Result rc = plGetSharedFontByType(&font, PlSharedFontType_Standard);
         if (R_SUCCEEDED(rc))
         {
-            Logger::info("Using Switch shared font");
-            Application::fontStash.regular = Application::loadFontFromMemory("regular", font.address, font.size, false);
+            Logger::info("Adding Switch shared standard font");
+            Application::fontStash.standard = Application::loadFontFromMemory("regular", font.address, font.size, false);
         }
 
-        // Korean font
-        rc = plGetSharedFontByType(&font, PlSharedFontType_KO);
-        if (R_SUCCEEDED(rc))
+        // Load other fonts on demand
+        bool isFullFallback = false;
+        AppletType at = appletGetAppletType();
+        if (at == AppletType_Application || at == AppletType_SystemApplication) // title takeover
         {
-            Logger::info("Adding Switch shared Korean font");
-            Application::fontStash.korean = Application::loadFontFromMemory("korean", font.address, font.size, false);
-            nvgAddFallbackFontId(Application::vg, Application::fontStash.regular, Application::fontStash.korean);
+            isFullFallback = true;
+            Logger::warning("Non applet mode, font full fallback is enabled!");
         }
+        
+        if (locale == 6 || locale == 15 || isFullFallback)
+        {
+            // S.Chinese font
+            rc = plGetSharedFontByType(&font, PlSharedFontType_ChineseSimplified);
+            if (R_SUCCEEDED(rc))
+            {
+                Logger::info("Adding Switch shared S.Chinese font");
+                Application::fontStash.schinese = Application::loadFontFromMemory("schinese", font.address, font.size, false);
+            }
+            // Ext S.Chinese font
+            rc = plGetSharedFontByType(&font, PlSharedFontType_ExtChineseSimplified);
+            if (R_SUCCEEDED(rc))
+            {
+                Logger::info("Adding Switch shared S.Chinese extended font");
+                Application::fontStash.extSchinese = Application::loadFontFromMemory("extSchinese", font.address, font.size, false);
+            }
+        }
+        if (locale == 11 || locale == 16 || isFullFallback)
+        {
+            // T.Chinese font
+            rc = plGetSharedFontByType(&font, PlSharedFontType_ChineseTraditional);
+            if (R_SUCCEEDED(rc))
+            {
+                Logger::info("Adding Switch shared T.Chinese font");
+                Application::fontStash.tchinese = Application::loadFontFromMemory("tchinese", font.address, font.size, false);
+            }
+        }
+        if (locale == 7 || isFullFallback)
+        {
+            // Korean font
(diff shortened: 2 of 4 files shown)


library/include/borealis/i18n.hpp
```diff
@@ -64,6 +64,14 @@ void loadTranslations();
  */
 std::string getCurrentLocale();
 
+#ifdef __SWITCH__
```
natinusala, line 67: Why put it here? Technically it's i18n related but it doesn't belong to the i18n system

library/lib/application.cpp
```diff
@@ -229,23 +229,126 @@ bool Application::init(std::string title, Style* style, LibraryViewsThemeVariant
...
+        // Load other fonts on demand
+        bool isFullFallback = false;
+        AppletType at = appletGetAppletType();
+        if (at == AppletType_Application || at == AppletType_SystemApplication) // title takeover
+        {
+            isFullFallback = true;
+            Logger::warning("Non applet mode, font full fallback is enabled!");
+        }
+        
+        if (locale == 6 || locale == 15 || isFullFallback)
```
natinusala, line 251: Not a fan of magic values here, could you define them somewhere instead?

library/lib/application.cpp
```diff
@@ -229,23 +229,126 @@ bool Application::init(std::string title, Style* style, LibraryViewsThemeVariant
...
-        // Korean font
-        rc = plGetSharedFontByType(&font, PlSharedFontType_KO);
-        if (R_SUCCEEDED(rc))
+        // Load other fonts on demand
+        bool isFullFallback = false;
+        AppletType at = appletGetAppletType();
+        if (at == AppletType_Application || at == AppletType_SystemApplication) // title takeover
+        {
+            isFullFallback = true;
+            Logger::warning("Non applet mode, font full fallback is enabled!");
```
natinusala, line 248: Is this really a warning?

library/include/borealis/i18n.hpp
natinusala: I mean it depends on the Switch implementation, and i18n is supposed to be generic (and eventually will)

library/lib/i18n.cpp
```diff
@@ -108,6 +108,34 @@ std::string getCurrentLocale()
...
+    if (R_SUCCEEDED(res))
+    {
+        if (R_SUCCEEDED(res))
+        {
+            res = setMakeLanguage(languageCode, &setlanguage);
+            return (int)setlanguage;
+        }
+        else
+        {
+            brls::Logger::error("Unable to convert system language ID (error 0x{0:x}), using the default one: {1}", res, DEFAULT_LOCALE);
```
natinusala, line 128: You can use `{0:#x}` instead of `0x{0x:}`

natinusala (changes requested):

natinusala:
Thanks for the PR!

`nvgAddFallbackFontId` works by appending the fallback, so the order would be regular then Korean then shared symbols.

I am going to wait for you to add proper names to the locale magic values since I am having a real hard time to understand the second part of the code without it.

library/lib/application.cpp
```diff
@@ -229,23 +229,126 @@ bool Application::init(std::string title, Style* style, LibraryViewsThemeVariant
...
+        int locale = i18n::swGetCurrentLocaleID();
 
         // Standard font
         Result rc = plGetSharedFontByType(&font, PlSharedFontType_Standard);
         if (R_SUCCEEDED(rc))
         {
-            Logger::info("Using Switch shared font");
-            Application::fontStash.regular = Application::loadFontFromMemory("regular", font.address, font.size, false);
+            Logger::info("Adding Switch shared standard font");
+            Application::fontStash.standard = Application::loadFontFromMemory("regular", font.address, font.size, false);
```
natinusala, line 239: Wouldn't that be `standard` as well in the `loadFontFromMemory` call?

natinusala (changes requested):

library/include/borealis/i18n.hpp
CaiMiao: this id only make sense for libnx i think
removed `#ifdef`

library/lib/application.cpp
CaiMiao: substituted with enforcing `SetLanguage` enum

library/lib/application.cpp
CaiMiao: changed to `INFO`

library/lib/i18n.cpp
CaiMiao: done

library/lib/application.cpp
CaiMiao: changed

natinusala:
I am very sorry but I am currently rewriting the entirety of font handling for the yoga rework (yoga branch, my work is not pushed yet).

I am currently adapting Behemoth's work of splitting borealis in two "platforms", glfw and switch. That means the Switch will have its own font loading code, which is no longer in `application.cpp`. The loading logic will be different too - everything will be abstracted away from the Switch definitions.

Would you mind redoing what you did on the yoga branch once I push what I have? I will notify you here.

The core of what you did will not be altered, you will just need to put it in the new font code dedicated to the Switch platform.

CaiMiao:
i'm okay with that

natinusala:
Hey @CaiMiao, I finished working on the new font abstraction layer. It's on the yoga branch, feel free to PR to that branch if you want to.

Now fonts are defined in a fully platform agnostic file, and every platform implementation loads them as they wish. The fallback code is still in `application.cpp` but the font loading has been moved to `glfw_font.cpp` and `switch_font.cpp`.

CaiMiao:
have problem with building demo
massive `<command-line>: warning: missing terminating " character`
and some errors deny me from building
```
<command-line>: error: missing terminating " character
<command-line>: note: in definition of macro 'BRLS_RESOURCES'
C:/src/msys64/home/CaiMiao/borealis/library/lib/core/font.cpp:25:29: note: in expansion of macro 'BRLS_ASSET'
   25 | #define MATERIAL_ICONS_PATH BRLS_ASSET("material/MaterialIcons-Regular.ttf")
      |                             ^~~~~~~~~~
C:/src/msys64/home/CaiMiao/borealis/library/lib/core/font.cpp:51:56: note: in expansion of macro 'MATERIAL_ICONS_PATH'
   51 |     return this->loadFontFromFile(FONT_MATERIAL_ICONS, MATERIAL_ICONS_PATH);
      |                                                        ^~~~~~~~~~~~~~~~~~~
make[1]: *** [/opt/devkitpro/devkitA64/base_rules:14：font.o] 错误 1
make: *** [Makefile:186：all] 错误 2
```

natinusala:
It looks like your `BRLS_RESOURCES` define is wrong, it's missing a quote

CaiMiao:
> It looks like your `BRLS_RESOURCES` define is wrong, it's missing a quote

I'm compiling the demo under the yoga branch, and everything is unmodified, so i don't really know where I did wrong.  
can only see `CXXFLAGS := $(CXXFLAGS) -DYG_ENABLE_EVENTS -fdata-sections -DBRLS_RESOURCES="\"romfs:/\""` in `borealis.mk`

natinusala:
Can you try again with the changes of this commit please?

https://github.com/natinusala/borealis/commit/6ea9f96f3366478eef746b332f63a35e8db020ac

CaiMiao:
> Can you try again with the changes of this commit please?
> 
> [6ea9f96](https://github.com/natinusala/borealis/commit/6ea9f96f3366478eef746b332f63a35e8db020ac)

this works for me, ty. once i tested the implement okay i will PR them again asap.

CaiMiao:
replaced by #79

CaiMiao:
well i think it is not bad to merge this to legacy though, in case anyone need to include the shared font workaround without need to update to yoga.

natinusala:
The thing is, I don't want to maintain this version anymore. The yoga rework is too hard to rebase since so many things changed 😕

PR #77, opened 2021-02-03
Touch control

XITRIX:
I've made a basic touch-event system, hope it could be useful.


natinusala:
Hey, thanks for the PR! I'm sorry I didn't get to review it the first time, seeing that it's pretty large.

It's going to take me some time to process everything, can you maybe explain quickly how your implementation works / how is its architecture? Thanks!

natinusala:
It does not look like your branch is building for Switch, I see glfw / OpenGL stuff being built 🙁

XITRIX:
That's strange... Just merged yoga branch and checked, make clean / make -j works fine.

natinusala:
You added glfw back to the Switch Makefile it seems

XITRIX:
Oh, I remember, in one of your old commits it was there, and I could not build nro, so I removed it by my self but not commited it, cause I thought you need it for any reason.
Is my last commit building?

natinusala:
I was able to try it by removing the glfw line in the Makefile.

It works quite well, good job! Don't forget to add your name to the copyrights if you didn't already 😄 

I don't know how far you wanted to go with your code, but there are a few things I noticed:
- when we use touch, the highlight doesn't disappear, while it does in HOS, it can be fixed by having a "touch" mode that's enabled when we use touch and disabled when we use the controller
- some gestures are a bit sloppy, especially scrolling, it doesn't accelerate when we flick the finger for instance. I think RetroArch has some pretty good acceleration code that's GPL so we can use it!
- pressing things only moves the highlight cursor, it doesn't actually "click" the thing. I think it should run the "A" (or "click") action if there's any on the clicked view. If you look on HOS, touching things with the finger actually clicks them.
- if you try to touch something that's not fully in view (so if it's clipped by scrolling), you won't be able to touch it, whereas in what you did it becomes focused and scrolls to the center. That's tied to the "fancy" scrolling that's not yet implemented in borealis so we can take care of it later I think.
- there's no audio but that's polishing for later 😛 

How do you want to proceed? Do you want me to look at the code and review it before changing anything? Or do you want to tackle some of these issues first?

natinusala:
In qlaunch when you touch on things, you need to touch them once to select them THEN touch them another time to click them. So actually borealis views should have two modes: a "touch to click" one and a "touch to highlight, touch another time to click" mode

XITRIX:
I think if you'll review it, it will be better cause I'm new in c++ (I'm c# developer). 
I tried to implement something similar like iOS handles touches, and I had a problem with idea of how acceleration on scroll should work so I'll try to find how it works in RetroArch. 
I know about current problems it have, cause I just used default giveFocus on items without any modifications. If you think it's ok, and you'd like me to continue touch development, I'll be happy to help with it :)

XITRIX:
Also I have no sounds in borealis at all, do you have any idea of why that could be?

natinusala:
Sound is only implemented on Switch right now, are you trying on Switch? It works for me, does it not work for you?

XITRIX:
Yes, I'm trying on switch with atmosphere 17.1, hos 11.1

natinusala:
Using touch or controller? Do you see anything relevant in the logs?

XITRIX:
Controller, I know that I've not implemented sound with touch. That what inside logs, I'm not sure what qlaunch is, may be it's some kind related:
image

natinusala:
Is your hbloader old by any chance?

XITRIX:
It's v3.4.0

XITRIX:
oh, it's hbmenu
I'll check right now, but I think it's latest one

HookedBehemoth:
This is a filesystem permission issue. The qlaunch romfs is accessible with hbloader since v2.3.3.

Don't use forwarders.

XITRIX:
I got it, I don't know why, but I cannot run hbmenu by holding R while running any title (albums works fine, but it's an applet), so I created a forwarder for hbmenu. It worked fine when I was on SXOS, but after updating on Atmosphere it stoped working, bo you know any reason of why that could happen?

natinusala:
It looks like a botched Atmos install and/or config to me 👀 

I'm going to read and review the changes when I can, it's a big PR so it might take me a while to take in lol

library/borealis.mk
```diff
@@ -8,6 +8,7 @@ include $(TOPDIR)/$(current_dir)/lib/extern/switch-libpulsar/deps.mk
 SOURCES		:=	$(SOURCES) \
 				$(current_dir)/lib/core \
 				$(current_dir)/lib/views \
+				$(current_dir)/lib/touch \
```
natinusala: I would rather have the code in core/touch instead

library/include/borealis/core/application.hpp
```diff
@@ -210,6 +210,11 @@ class Application
      */
     static std::string getLocale();
 
+    /**
+     * View that was recognized as target for touch events.
+     */
+    inline static View* firstResponder;
```
natinusala: What is the purpose of storing this in the global app state like so? I would like to avoid doing that as much as possible

library/include/borealis/core/application.hpp
```diff
@@ -234,6 +239,7 @@ class Application
 
     inline static ControllerState oldControllerState = {};
     inline static ControllerState controllerState    = {};
+    inline static TouchState touchState              = {};
```
natinusala: You actually showed me something here: frame state should not be stored in the global app state.

Here, controllerState and touchState are the state of the _current frame_, they do not belong in Application.

I'm going to remove the controller state, can you do the same for touch?

library/include/borealis/core/platform.hpp
```diff
@@ -94,7 +95,13 @@ class Platform
     virtual InputManager* getInputManager() = 0;
 
     /**
-     * Returns the FontLoader for the platform.
+     * Returns the TouchManager for the platform.
+     * Cannot return nullptr.
+     */
+    virtual TouchManager* getTouchManager() = 0;
```
natinusala: Actually I intended to have touch part of the InputManager, seeing that the code of each impl is quite short (just one method).

Can you move updateTouchState to InputManager and remove TouchManager?

library/include/borealis/core/touch.hpp
```diff
@@ -0,0 +1,49 @@
+/*
+    Borealis, a Nintendo Switch UI Library
+    Copyright (C) 2021  natinusala
```
natinusala: Why is my copyright here? :p

library/include/borealis/core/touch.hpp
```diff
@@ -0,0 +1,49 @@
...
+    START,
+    STAY,
+    END,
+    NONE,
+};
+
+typedef struct TouchState
+{
+    TouchEvent state = TouchEvent::NONE;
+    double x, y;
```
natinusala: Why is state initialized and not x, y?

Why are x, y double while the brls coordinates system uses float for everything?

natinusala (changes requested):
Here's a first batch of comments, I really like what you did so far gz!

I have a question, does the touch code take scaling into account? The window size does not always match the view coordinates system size. The view coordinates are always as close as possible to 1080x720, and if the actual window is bigger the views are scaled up on display, but the coordinates system stays the same.

library/include/borealis/core/application.hpp
XITRIX: I search it inside Application once first touch was made and need to store it until touch will be ended.
I send all touch inputs to it and also use it inside View to interrupt other recognisers if one change its state from UNSURE to START.
I could make it private and throw it as argument from Application but I don't think it's an elegant solution.

XITRIX:
I converted window coordinates to app's one. I tested it on my laptop with cursor, I scaled window in different ways and everything worked fine.

library/include/borealis/core/touch.hpp
XITRIX: It's your lib, so there is your copyright also :)

library/include/borealis/core/touch.hpp
XITRIX: That's cause glfwGetCursorPos gets double as arguments and so I used double everywhere. I'll change it.

library/include/borealis/core/application.hpp
XITRIX: I'm not sure where it should be, could you remove the controllerState first and I'll copy your solution?

library/borealis.mk
XITRIX: done

library/include/borealis/core/platform.hpp
XITRIX: done

library/include/borealis/core/touch.hpp
XITRIX: done

library/include/borealis/core/application.hpp
natinusala: Do you think it would be possible to have this inside FrameContext instead, and toss the context around where you need it? Would it make touch eventss more difficult to use?

library/include/borealis/core/application.hpp
natinusala: I will put it in FrameContext yes, I'll notify you when I'm done

library/include/borealis/core/touch.hpp
natinusala: Yeah the brls coordinates use float, I'd rather have the main code use the standard stuff and have the platform code adapt if needed (so in that case, convert from double to float).

library/include/borealis/core/application.hpp
XITRIX: As I can see FrameContext constructs after all input logic done, do you want to move it inside Application::frame() ?
I think it will not be a problem, I'll just add a firstResponder argument in gestureRecognizerRequest method, and will toss it from Application where we construct FrameContext, cause I don't think that I need full FrameContext inside touch logic. 
Just notify me when you'll done these changes, I'll adopt my code to your solution.

XITRIX:
Will you add an horizontal scroll any time soon? I would like to test touch inputs with complex view hierarchy like an array of touchable images inside horizontal scroll inside vertical scroll with some text in it, like in eShop app page.

XITRIX:
How did you took the names of sounds in qlaunch.bfsar?
I spend several hours already and still cannot understand how to do that...

natinusala:
Sorry for not responding earlier, I'm in the process of replacing some files to relicense the library. I want to do that before accepting more contributions, and so I'm going to take care of the frame context stuff later, when the relicense is done.

> Will you add an horizontal scroll any time soon?

Soon I don't know, but I'm eventually going to need it, if you can you can show me what you have and I can comment / help out.

Keep in mind that for lists of repeated items it's better to use a recycling mechanism, and that's not yet implemented (but you can probably do it yourself).

> How did you took the names of sounds from qlaunch.bfsar?

I believe the pulsar library has a function to list sound names of an archive.

XITRIX:
> I believe the pulsar library has a function to list sound names of an archive.
I had not found it, but finally I've made it myself, just ran through lots of ids and list their names if it exists.

Ok, may be I'll try to make them myself, when I'll finish with touch.

natinusala:
I will look at everything you did in a giffy.

I have to warn you that the license of the library has been changed from GPLv3 to Apache 2.0, if you don't agree to that change please close the PR. Otherwise you can add your name to the authors file if you haven't already (and resolve the conflicts 👀 ).

natinusala:
FYI I just pushed a change that removes controllerState from the global Application state

library/include/borealis/core/application.hpp
XITRIX: Removed from public header

library/include/borealis/core/application.hpp
XITRIX: Moved first responder from public header.

XITRIX:
> I have to warn you that the license of the library has been changed from GPLv3 to Apache 2.0

I don't even know what does that mean, so it is not a problem for me :)

demo/captioned_image.cpp
```diff
@@ -42,6 +42,11 @@ CaptionedImage::CaptionedImage()
     this->forwardXMLAttribute("imageHeight", this->image, "height");
 
     this->forwardXMLAttribute("caption", this->label, "text");
+
+    this->addGestureRecognizer(new brls::TapGestureRecognizer([this](auto _){
```
natinusala: I don't really like usage of auto especially in that case where there is no clue as to what is the function argument here (no type, no name)

library/include/borealis/core/application.hpp
```diff
@@ -204,6 +205,17 @@ class Application
      * Returns the current system locale.
      */
     static std::string getLocale();
+    
+    /**
+     * If touch input mode enabled, disable it and move focus on last view
+     * Returns true if touch mode was disabled.
+     */
+    static bool dismissTouchMode();
```
natinusala: Actually, for the sake of extensibility, I wonder if it would not be better to use an enum here instead of a boolean. I'm thinking of an eventual "mouse mode" that would work yet again differently from both touch and controller mode.

library/include/borealis/core/application.hpp
```diff
@@ -226,6 +238,8 @@ class Application
 
     inline static View* currentFocus;
 
+    inline static bool focusTouchMode = false;
```
natinusala: My comment above was more for that line actually, sorry about that

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <functional>
+
+#include <borealis/core/input.hpp>
+
+namespace brls
+{
+
+class View;
+
+enum class GestureState
```
natinusala: Can you add a description for every enum entry please?

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,50 @@
...
+{
+    INTERRUPTED,
+    UNSURE,
+    START,
+    STAY,
+    END,
+    FAILED,
+};
+
+class GestureRecognizer
```
natinusala: Can you add documentation for the class as well as for every public method please?

natinusala (changes requested):

natinusala:
Not that I doubt of the answer but I have to ask for big contributions like that, the code is from you right? It doesn't come from any external project, or even inspired from (code washing) external projects, or their design?

XITRIX:
Sorry, I've missed that question.
I have written that code for that project from scratch. I used only concept of how it works in iOS, but most of it is just my idea of how it works. I used some names from real iOS functions, but it's only names.

demo/captioned_image.cpp
XITRIX: I used it like that cause it is useless in that case, but ok, I'll replace it.

library/include/borealis/core/touch/gesture_recognizer.hpp
XITRIX: I'm not good at documentation, but I've done my best. Hope it's ok.

demo/captioned_image.cpp
natinusala: It's useless I agree but I never used auto anywhere in the project and I don't intend to. There are a few places where auto is used from code by other contributors, I will probably go through all of them and remove them.

library/include/borealis/core/application.hpp
```diff
@@ -54,6 +55,11 @@ namespace brls
     void frame(FrameContext* ctx) override;
 }; TODO: restore that */
 
+enum class InputType {
```
natinusala: Can you add some documentation to explain what it is? 🙈

I know what it is because I suggested to add it, but the name is kinda generic and I feel like people won't know what purpose it serves.

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+// Superclass for all the other recognizers
+class GestureRecognizer
+{
+public:
+    // Main recognition loop, for internal usage only, should not be called anywhere, but Application
+    virtual GestureState recognitionLoop(TouchState touch, View* view);
+    
+    // Returns true if recognizer plays default 'click' sound on touch start 
+    virtual bool soundOnTouch();
```
natinusala: I don't like storing the "should play a sound" state in a class attribute and have it set in recognitionLoop then returned in soundOnTouch.

IMO the "should a sound be played" boolean should be returned directly from recognitionLoop, and not carried around in the recognizer state like that.

I think it's going to be hard to return the boolean from recognitionLoop since you already return a GestureState, but it could be passed as a pointer?

`virtual GestureState recognitionLoop(TouchState touch, View* view, bool* shouldPlaySound);`

This whole sound situation could be improved a bit more actually, since we need to play the "nothing to touch here" sound when the user touches somewhere on the screen that doesn't have anything. A boolean might not be enough to represent all of our use cases.

So maybe instead of a "should play sound" boolean, we should add a "touch state" value that represents the gesture state: did it hit? did it miss? You go from view to view gathering the state, and then you play the sound in Application based on the final state.

I don't know if I'm really clear, let me know if you see what I mean

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,63 @@
...
+namespace brls
+{
+
+class View;
+
+// Represents current gesture state
+enum class GestureState
+{
+    INTERRUPTED, // Gesture has been interupted, no callbacks will come
+    UNSURE, // Gesture started recognition and not sure if it should interupt other recognizers
```
natinusala: I don't find the name UNSURE alone very self explanatory, how about something like START_UNSURE_INTERRUPT?

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,63 @@
...
+    // Returns true if recognizer plays default 'click' sound on touch start 
+    virtual bool soundOnTouch();
+    
+    // Interupt this recognizer
+    // If onlyIfUnsureState == true recognizer will be interupted
+    // only if current state is UNSURE
+    void interrupt(bool onlyIfUnsureState);
+    
+    // If false, this recognizer will be skipped
+    bool enabled = true;
```
natinusala: Make it private and add a getter / setter pls

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,63 @@
...
+    
+    // Get the current state of recognizer
+	GestureState getState() const { return state; }
+protected:
+	GestureState state;
```
natinusala: Uninitialized attribute

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,63 @@
...
+};
+
+};
```
natinusala: No need to add a semicolon after the namespace, also add `// namespace brls` pls

library/include/borealis/core/touch/gesture_recognizer.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <borealis/core/input.hpp>
+
+namespace brls
+{
+
+class View;
+
+// Represents current gesture state
+enum class GestureState
```
natinusala: I think the base gesture recognized interface should be in "core/gesture.hpp" since it's part of the library "core" (users can implement their own), and then every recognizer can be in core/touch/xxx

library/include/borealis/core/touch/pan_gesture_recognizer.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+#pragma once
+
+#include "gesture_recognizer.hpp"
+
+namespace brls
+{
+
+class PanGestureRecognizer;
+typedef std::function<void(PanGestureRecognizer*)> PanGestureRespond;
```
natinusala: Do you think it would be possible to change the signature of the listener to have everything passed as an individual parameter instead of giving the huge recognizer? Like `typedef std::function<void(float y, float acceleration...)> PanGestureRespond;`

Or if it's too boilerplate, maybe create a struct "PanGestureState" that holds everything and give a pointer to that instead?

natinusala (changes requested):
You should run clang-format with `./scripts/format.sh --fix`, it will take care of all the formating for you

library/include/borealis/core/touch/gesture_recognizer.hpp
XITRIX: I think it's quite self explanatory, i.e. I just hold my finger on screen and recognizer is UNSURE what I want to do. I mean it can't tell does it should fail or start tracking that gesture, so it is UNSURE. It is not about, is it sure that it should stop other recognizers.

library/include/borealis/core/touch/pan_gesture_recognizer.hpp
XITRIX: I'm not sure about changing it. I think having full recogniser in callback will give us more possibilities to control it, i.e. may be we'd like to disable it when it's done. Also I think having a separate struct in callback is just duplication of already existing data, it's much easier to make some fields in recogniser as public and allow to take them. What do you think about it?

library/include/borealis/core/touch/gesture_recognizer.hpp
XITRIX: I've made some changes in recognition loop, check it. I am fully agreed that something should be changed with "sound for nothing", but right now I have no any idea of how I could improve it...

"touch state" could be difficult to use, cause there are some cases where it's useless, i.e. pan recognizer, it's ok with touch, but it's still should play "nothing" sound, or tap recognizer should not play it, cause it has it's own "on pressed" sound, but if button is disabled, or has no actions on it, it should play "nothing" sound anyway.  

So I decided to use bool, that could be calculated in runtime and if no one recognizer changed this value to false, so app can play sound.

library/include/borealis/core/touch/pan_gesture_recognizer.hpp
natinusala: I feel like if you need the recognizer in the callback you should capture it in the lambda. The callback is for "state change" event and so it should only give new state data...

Speaking of which, would it be possible to use brls::Event instead of a "raw" std::function?

library/include/borealis/core/platform.hpp
```diff
@@ -92,7 +92,7 @@ class Platform
     virtual InputManager* getInputManager() = 0;
 
     /**
-     * Returns the FontLoader for the platform.
+ 	 * Returns the FontLoader for the platform.
```
natinusala: wrong indent

library/include/borealis/core/touch/gesture_recognizer.hpp
natinusala: How about replacing the bool with the sound to play? And then use `brls::SOUND_NONE` if you don't want to play anything.

library/include/borealis/core/touch/pan_gesture.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+class PanGestureRecognizer;
+typedef std::function<void(PanGestureRecognizer*)> PanGestureRespond;
+
+// Axis of pan recognition start conditions
+enum class PanAxis
+{
+    HORIZONTAL, // Triggers only on horizontal coordinate changes
+    VERTICAL, // Triggers only on vertical coordinate changes
+    NONE, // Any movement allowed
```
natinusala: "NONE" would mean no axis, wouldn't it be better to call it "ANY" or "ALL"?

library/include/borealis/core/touch/pan_gesture.hpp
```diff
@@ -0,0 +1,103 @@
...
+};
+
+struct position
```
natinusala: That struct is exposed to everyone as brls::position, I would rename it to PanPosition

library/include/borealis/core/touch/pan_gesture.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+    float x;
+    float y;
+};
+
+// Contains info about acceleration on pan ends
+struct pan_acceleration
```
natinusala: PanAcceleration

library/include/borealis.hpp
```diff
@@ -52,3 +53,7 @@
 #include <borealis/views/scrolling_frame.hpp>
 #include <borealis/views/sidebar.hpp>
 #include <borealis/views/tab_frame.hpp>
+
+//Gestures
```
natinusala: `// Gestures`

library/include/borealis/core/gesture.hpp
```diff
@@ -0,0 +1,64 @@
...
+};
+
+// Superclass for all the other recognizers
+class GestureRecognizer
+{
+  public:
+    // Main recognition loop, for internal usage only, should not be called anywhere, but Application
+    virtual GestureState recognitionLoop(TouchState touch, View* view, bool* shouldPlayDefaultSound);
+
+    // Interupt this recognizer
```
natinusala: `Interrupt`

library/include/borealis/core/gesture.hpp
```diff
@@ -0,0 +1,64 @@
...
+class View;
+
+// Represents current gesture state
+enum class GestureState
+{
+    INTERRUPTED, // Gesture has been interupted, no callbacks will come
+    UNSURE, // Gesture started recognition and not sure if it should interupt other recognizers
+    START, // Gesture sure that it match to conditions and will interupt other recognizers
+    STAY, // Gesture in process, user still hold finger on screen
+    END, // User released it's finger from screen, final frame of gesture
```
natinusala: their finger*

library/lib/platforms/switch/switch_input.cpp
```diff
@@ -66,4 +67,48 @@ void SwitchInputManager::updateControllerState(ControllerState* state)
     }
 }
 
+void SwitchInputManager::updateTouchState(TouchState* state)
+{
+    // Get gamepad state
+    float x = oldTouch.x;
+    float y = oldTouch.y;
+
+    HidTouchScreenState hidState = { 0 };
```
natinusala: Make it static to prevent it from being pushed to the stack every frame (unless the compiler already optimizes it?)

library/include/borealis/platforms/glfw/glfw_input.hpp
```diff
@@ -32,8 +32,11 @@ class GLFWInputManager : public InputManager
 
     void updateControllerState(ControllerState* state) override;
 
+    void updateTouchState(TouchState* state) override;
+
   private:
     GLFWwindow* window;
+    TouchState oldTouch;
```
natinusala: It can me made local and static to GLFWPlatform::updateTouchState

library/lib/platforms/glfw/glfw_input.cpp
```diff
@@ -110,4 +112,39 @@ void GLFWInputManager::updateControllerState(ControllerState* state)
...
+    double x, y;
+    glfwGetCursorPos(this->window, &x, &y);
+
+    state->x = x / Application::windowScale;
+    state->y = y / Application::windowScale;
+
+    int clickState = glfwGetMouseButton(window, GLFW_MOUSE_BUTTON_LEFT);
+    if (clickState == GLFW_PRESS)
+    {
+        if (oldTouch.state == TouchEvent::START || oldTouch.state == TouchEvent::STAY)
```
natinusala: Actually this whole oldTouch logic can be factorized in Application, right now it's duplicated between GLFW and Switch
 
The whole TouchEvent should be moved out of the input manager and automated. The input manager should say that "x fingers are on screen at those positions at the current frame" and Application should automatically deduce the "events" by counting the fingers and seing when a finger was removed or placed on screen compared to previous frame.

The notion of "event" is too high level for a platform driver, and it makes for duplicate logic.

natinusala (changes requested):

library/include/borealis/core/touch/pan_gesture.hpp
XITRIX: What do you think about implementing structures for view's frame? 
i.e. instead of 4 vars (x, y, width, height) create something like:

struct Point { x, y }
struct Size { width, height }
struct Rect { Point, Size }

and use Rect to store all those vars.

I think they could be useful, also I could use one here.

library/include/borealis/core/touch/pan_gesture_recognizer.hpp
XITRIX: I changed whole recognizer to status structs.
About brls::Event, I don't think that it's what we should use with recognizers, I feel like recognizer should call one and only one function which could include invoking Events inside it, but not contain an event subscriber in recognizer it self. I can't think of a single example where having several callbacks on recognizer could be useful. Anyway, if you think it should have this functionality, that's not a problem, I'll add it.

library/include/borealis/core/touch/pan_gesture.hpp
natinusala: The thing is that a view does not have a width or height stored, it's stored internally in the yoga node and returned by `YGNodeLayoutGetWidth` when we need it.

Unless you changed that and I haven't got to that point in reading the PR yet?

Sorry for the long delay I was in vacation 🌴 

I will resume the review in a few days after wrapping some other things up

XITRIX: Hope you had a good rest!

No I haven't changed it yet. I've made another brunch where I tried to implement that idea and I found that it's yoga's node variables :\ 

But still I think we could change lot's of: 
float originX; 
float originY; 

to one:
Point origin;

I think those structs will be useful for lots of another things. When I'll finish with that idea, I'll create a PR for it.

natinusala: I mean yeah, I want to avoid boilerplate though, how many times and where do you think we could use that new struct in the code?

XITRIX: I'm not sure about Size and Rect, but Point could be used everywhere: View's detachedOrigin, translation, Touch\Mouse position, velocity, and lots of other stuff, which contains X and Y variables. These examples is what first came to mind.

natinusala: Go for it then I guess

XITRIX: Should I do that in separate brunch, or I can do it here?

natinusala: You can do it here 👍

library/lib/platforms/glfw/glfw_input.cpp
XITRIX: Moved that part to base InputManagaer

natinusala:
@XITRIX do you plan to add anything else to the `metrics` folder?

I wonder if, for the sake of simplicity, we shouldn't move it to `core/geometry.hpp` with everything inside? And a `core/geometry.cpp` file with the code, I want to avoid header-only code as much as possible now.

XITRIX:
Ok, not a problem, I'll move it.

library/include/borealis/core/geometry.hpp
```diff
@@ -0,0 +1,107 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#pragma once
+
+#include <String>
```
natinusala: uppercase String? :thinking:

library/include/borealis/core/gesture.hpp
```diff
@@ -0,0 +1,64 @@
...
+    INTERRUPTED, // Gesture has been interupted, no callbacks will come
+    UNSURE, // Gesture started recognition and not sure if it should interupt other recognizers
+    START, // Gesture sure that it match to conditions and will interupt other recognizers
+    STAY, // Gesture in process, user still hold finger on screen
+    END, // User released their finger from screen, final frame of gesture
+    FAILED, // Gesture failed conditions
+};
+
+// Superclass for all the other recognizers
+class GestureRecognizer
```
natinusala, line 67: Missing virtual dtor

natinusala: `warning: ‘class brls::GestureRecognizer’ has virtual functions and accessible non-virtual destructor`

library/include/borealis/core/input.hpp
```diff
@@ -66,11 +69,34 @@ enum ControllerAxis
...
+enum class TouchPhase
+{
+    START,
+    STAY,
+    END,
+    NONE,
+};
+
+// Contains raw touch data
+struct RawTouch
```
natinusala: I'm not sure about the name of the two structs here, "touch" does not represent anything

RawTouch could be named RawTouchState since it represents the state of touch in the lower level (platform driver).

 And so Touch could be named TouchState

library/include/borealis/core/touch/pan_gesture.hpp
```diff
@@ -0,0 +1,85 @@
...
+    Point position; // Current position
+    Point startPosition; // Start X position
+    Point delta; // Difference between current and previous positions by X
+
+    // Acceleration info, NOT NULL ONLY from
+    // gesture callback and when current state is END
+    PanAcceleration acceleration;
+};
+
+typedef std::function<void(PanGestureStatus)> PanGestureRespond;
```
natinusala: Why don't you use a brls::Event here instead of the std::function?

library/include/borealis/core/gesture.hpp
```diff
@@ -0,0 +1,64 @@
...
+{
+  public:
+    // Main recognition loop, for internal usage only, should not be called anywhere, but Application
```
natinusala, line 72: As people could theoretically implement their own gesture recognizers, I think this method should have a little bit more documentation as to what it should do and what the parameters are

library/include/borealis/core/touch/tap_gesture.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace brls
+{
+
+struct TapGestureStatus
+{
+    GestureState state; // Gesture state
+    Point position; // Current position
+};
+typedef std::function<bool(TapGestureStatus)> TapGestureRespond;
```
natinusala: Same here, a brls::Event could be used

library/include/borealis/core/view.hpp
```diff
@@ -1032,6 +1034,35 @@ class View
...
+    void addGestureRecognizer(GestureRecognizer* recognizer);
+
+    /**
+     * Called each frame when touch is registered.
+     * 
+     * @return false if there is no recognizer 
+     * that needs to play it's own click sound, 
+     * so play default is allowed
+     */
+    bool gestureRecognizerRequest(Touch touch, View* firstResponder);
```
natinusala: I feel the same about the "return false to play a sound", it should be passed as a parameter, maybe a "sound action" enum like PLAY_DEFAULT, DONT_PLAY, etc...

library/include/borealis/platforms/switch/switch_input.hpp
```diff
@@ -1,5 +1,6 @@
 /*
     Copyright 2021 natinusala
+	Copyright (C) 2021  XITRIX
```
natinusala: Wrong indent

library/lib/core/application.cpp
```diff
@@ -161,10 +163,45 @@ bool Application::mainLoop()
...
+
+    // Touch controller events
+    switch (touchState.phase)
+    {
+        case TouchPhase::START:
+            Logger::debug("Touched at X: " + std::to_string(touchState.position.x) + ", Y: " + std::to_string(touchState.position.y));
+            Application::setInputType(InputType::TOUCH);
+
+            // Search for first responder, which will be the root of recognition tree
+            firstResponder = Application::activitiesStack[Application::activitiesStack.size() - 1]
```
natinusala: That will crash if the activity does not have a content view

library/lib/core/application.cpp
```diff
@@ -161,10 +163,45 @@ bool Application::mainLoop()
...
+            break;
+    }
+
+    if (firstResponder)
+    {
+        if (firstResponder->gestureRecognizerRequest(touchState, firstResponder))
+        {
+            // Play touch sound with random pitch
+            float pitch = (rand() % 10) / 10.0f + 1.0f;
+            Application::getAudioPlayer()->play(SOUND_TOUCH, pitch);
```
natinusala: There are multiple sounds to play depending on the view, just like when using a controller

For instance the clicking sound on a sidebar item is different than the clicking sound on a button. So each view should provide a "sound to play when touched", maybe depending on the touch event too

You should open the HOS settings app and touch it everywhere to see how it behaves, it will be better than my poor explanations lol

library/lib/core/application.cpp
```diff
@@ -317,13 +358,32 @@ void Application::onControllerButtonPressed(enum ControllerButton button, bool r
...
 
+bool Application::setInputType(InputType type)
+{
+    if (type == Application::inputType)
+        return false;
+
+    Application::inputType = type;
+
+    if (type == InputType::GAMEPAD)
+    {
```
natinusala: No need for braces

library/lib/core/application.cpp
```diff
@@ -221,6 +258,10 @@ void Application::quit()
 
 void Application::navigate(FocusDirection direction)
 {
+    // Dismiss navigation if input type was changed
```
natinusala, line 274: Is it actually like that in HOS?

library/lib/core/application.cpp
```diff
@@ -317,13 +358,32 @@ void Application::onControllerButtonPressed(enum ControllerButton button, bool r
...
+}
+
 View* Application::getCurrentFocus()
 {
     return Application::currentFocus;
 }
 
 bool Application::handleAction(char button)
 {
+    // Dismiss if input type was changed
```
natinusala, line 394: Same question here, is it actually like this in HOS?

library/lib/core/application.cpp
```diff
@@ -345,6 +405,7 @@ bool Application::handleAction(char button)
 
             if (action.available)
             {
+
```
natinusala: :eyes:

library/lib/core/box.cpp
```diff
@@ -316,6 +316,27 @@ View* Box::getDefaultFocus()
     return nullptr;
 }
 
+View* Box::hitTest(Point point)
+{
+    // Check if touch fits in view frame
+    if (getFrame().pointInside(point))
```
natinusala: Missing this->

library/lib/core/geometry.cpp
```diff
@@ -0,0 +1,142 @@
+/*
+    Copyright 2021 natinusala
```
natinusala: Wrong copyright :p

library/lib/core/input.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace brls
+{
+
+Touch InputManager::computeTouchState(RawTouch currentTouch, Touch lastFrameState)
+{
+    if (currentTouch.pressed)
+    {
+        lastFrameState.position = currentTouch.position;
+        if (lastFrameState.phase == TouchPhase::START || lastFrameState.phase == TouchPhase::STAY)
```
natinusala, line 27: No need for braces around single-line if/else

library/lib/core/touch/pan_gesture.cpp
```diff
@@ -0,0 +1,149 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <borealis.hpp>
+
+#define FPS 60.0f // TODO: get real FPS
```
natinusala: Will be fixed by #154

natinusala (changes requested):

library/lib/core/application.cpp
XITRIX: Yes, if view was selected by touch, first click on any navigation button will show the frame around that view without moving frame to next view.

library/lib/core/application.cpp
XITRIX: Yes, it is.

EmreTech:
Looking at the files, I noticed that the scrolling example text (in resources/i18n/en-US/scroll.json) is super long and is in Russian. I'm curious, is there a reason for this? You could make it shorter, like a lot shorter.

natinusala:
Tell me how you would test scrolling without a super long text?

XITRIX:
+1, I used it to test long scrolling view. True, it's on Russia, but I could replace it with lorem ipsum text. It was the first I found in google with "super long text" request.

Also I've found, that there is the problem with cyrillic character "Ц" in glfw version of font, it shows like " ," without "U".

EmreTech:
It's understandable if the super long text is for the scrolling, but yeah the text should be changed to the lorem ipsum example.

You could also move the current example to another folder for Russian text.

EmreTech:
> Tell me how you would test scrolling without a super long text?

Well, I thought an alternative could be a bunch of elements that you scroll down on. For example, a bunch of pictures on the screen, maybe some big buttons that do nothing, rectangles? And yes, I'm aware my example is similar to the first tab in the demo.

library/lib/core/application.cpp
XITRIX: I think you don't get what that sound is. This sound is "nothing to touch", it's the same everywhere you click on unclickable part of UI. 

I attached screenshots where you can see the parts (not strikethrough) where this sound generated, if you'll check, I fully reproduced HOS sound behaviour, where selected sideview tab not selecting again and produce that "nothing" sound, or disabled\no action button also produce that sound.

IMG_1151-min
IMG_7745-min

library/lib/core/input.cpp
XITRIX: Bad habit from swift :p (no if\else without braces)

XITRIX:
I've changed sound processing a bit. Now I'm not calling play() inside recognizer's callback anymore, now I'm setting sound to the pointer, provided by recognitionLoop() and play it inside Application. If there are several recognizers with custom sound, only the closest to the firstFesponder will be taken.

EmreTech:
I wanted to try this out, and I like it. It controls pretty smoothly, even on a Mac. Only one suggestion (might be hard to implement), I wish the scrolling on GLFW platforms was easier to do (what I mean is adding support for the scroll-wheel).

XITRIX:
I thought a lot about scrolling by wheel but I cannot realise how I could do that without any problem for users of Borealis...
The main idea is to allow pan gesture recognizer work with both scenarious, like it works on iOS apps ported on Mac, it's gestures automatically translates to Mac's equivalents.

natinusala:
Borealis is not about mouse, it's about controllers and touch. As mouse is
yet another "input type" with different control schemes, I wouldn't focus
on making mouse usable for now.

XITRIX:
I've added mouse scrolling, tried to use as less code as possible. Review it please.

natinusala:
Is it implemented as part of touch or separately? Because scrolling using the wheel doesn't make sense in a touch context, since... touch screens don't have a mouse wheel. If it works it's fine but it's weird to have a hybrid touch emulation / scroll wheel on a computer.

I would rather implememt mouse as mouse with its own defined behavior and stuff

XITRIX:
As you could guess, I'm using ios and macos as reference. Apple's UIKit has auto adaptation of its gesture recognizers for touch and mouse, so I can use the same recognizer on both platforms. I wanted to implement something like that, but right now I've implemented only scrolling wheel as separate part for mouse. May be I should fully separate their inputs and combine them in single struct i.e. "ScreenInput" and toss it through recognizers, but that will double the code of recognizers, cause then we'll need to handle both touch and mouse. I'm still not sure what will be the best solution, so decided to write something what will not have too many lines of code and will be easy to replace later.

EmreTech:
Wow, it works really fluently! It feels like i'm using a web browser, which is good.

natinusala:
@XITRIX And what purpose does it serve at the moment? Does mouse still act as a touch pointer on PC? Isn't it weird to have a half touch half mouse thing?

XITRIX:
You could test it, it works perfectly fine now, like it should work on PC.

library/include/borealis/core/audio.hpp
```diff
@@ -30,6 +30,8 @@ enum Sound
     SOUND_CLICK_ERROR, // played when the user clicks a disabled button / a view focused with no click action
     SOUND_HONK, // honk
     SOUND_CLICK_SIDEBAR, // played when a sidebar item is clicked
+    SOUND_TOUCH_UNFOCUS, // played when touch focus has been interrupted
+    SOUND_TOUCH, // played when touch not requires it's own click sound
```
natinusala: doesn't require*

library/include/borealis/core/input.hpp
```diff
@@ -66,11 +69,36 @@ enum ControllerAxis
...
+// Represents a touch phase in the current frame
+enum class TouchPhase
+{
+    START,
+    STAY,
+    END,
+    NONE,
+};
+
+// Contains raw touch data
```
natinusala: Maybe specify the difference here between the two structs?

Something like `Contains raw touch data, filled in by input platform code` and `Contains touch data automatically filled with current phase by the library`

library/include/borealis/core/touch/scroll_gesture.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#pragma once
+
+#include <borealis/core/touch/pan_gesture.hpp>
+
+namespace brls
+{
+
+/*
+ * Scroll recognizer
```
natinusala, line 25: I'm fine with having the scrolling code in the touch code as long as there is a TODO that says to move it somewhere else if mouse gets properly implemented. Can you add it?

When this PR gets merged I'll replace the TODO by a proper issue on GH.

library/include/borealis/core/touch/scroll_gesture.hpp
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Child of Pan recognizer.
+ * The only difference is that Scroll recognizer will translate scrolling wheel to pan gestures.
+ * If NO_TOUCH_SCROLLING=true defined, will ignore scrolling by touch.
```
natinusala, line 29: Shouldn't the "should touch scroll" boolean be defined in the platform instead of as a define?

library/include/borealis/core/touch/tap_gesture.hpp
```diff
@@ -0,0 +1,80 @@
...
+    // Get tap gesture event
+    TapGestureEvent getPanGestureEvent() const { return tapEvent; }
+
+  private:
+    TapGestureEvent tapEvent;
+    Point position;
+    GestureState lastState;
+};
+
+};
```
natinusala: No need for semicolon, add `// namespace brls`

library/include/borealis/core/touch/scroll_gesture.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * Also PanGestureStatus will be returned with delta values ONLY.
+ */
+class ScrollGestureRecognizer : public PanGestureRecognizer
+{
+  public:
+    ScrollGestureRecognizer(PanGestureEvent::Callback respond, PanAxis axis);
+    GestureState recognitionLoop(TouchState touch, View* view, Sound* soundToPlay) override;
+};
+
+};
```
natinusala: Same here

library/include/borealis/core/view.hpp
```diff
@@ -1040,6 +1042,35 @@ class View
...
+
+    /**
+     * Add new gesture recognizer on this view.
+     */
+    void addGestureRecognizer(GestureRecognizer* recognizer);
+
+    /**
+     * Called each frame when touch is registered.
+     * 
+     * @return false if there is no recognizer
```
natinusala: Outdated comment, it doesn't return a boolean anymore

library/include/borealis/platforms/switch/switch_input.hpp
```diff
@@ -31,8 +32,11 @@ class SwitchInputManager : public InputManager
 
     void updateControllerState(ControllerState* state);
 
+    void updateTouchState(RawTouchState* state);
+
   private:
     PadState padState;
+    TouchState oldTouch;
```
natinusala: Not necessary anymore

library/lib/core/application.cpp
```diff
@@ -149,6 +150,7 @@ void Application::createWindow(std::string windowTitle)
 bool Application::mainLoop()
 {
     static ControllerState oldControllerState = {};
+    static View* firstResponder;
```
natinusala, line 153: `touchFirstResponder` ?

library/lib/core/application.cpp
```diff
@@ -159,10 +161,60 @@ bool Application::mainLoop()
...
 
+    static TouchState oldTouch;
+    TouchState touchState = InputManager::computeTouchState(rawTouch, oldTouch);
+    oldTouch              = touchState;
+
+    // Touch controller events
+    switch (touchState.phase)
+    {
+        case TouchPhase::START:
+            Logger::debug("Touched at X: " + std::to_string(touchState.position.x) + ", Y: " + std::to_string(touchState.position.y));
```
natinusala, line 178: We have to remember to remove all debug prints once we're done

library/lib/core/application.cpp
```diff
@@ -159,10 +161,60 @@ bool Application::mainLoop()
...
+            firstResponder = Application::activitiesStack[Application::activitiesStack.size() - 1]
+                                 ->getContentView()
+                                 ->hitTest(touchState.position);
+    }
+
+    if (firstResponder)
+    {
+        Sound sound = firstResponder->gestureRecognizerRequest(touchState, firstResponder);
+        float pitch = 1;
+        if (sound == SOUND_TOUCH)
```
natinusala, line 210: IIRC in HOS there are more sounds that have their pitch randomized, did you poke around to see for yourself?

library/lib/core/touch/pan_gesture.cpp
```diff
@@ -0,0 +1,149 @@
...
+#define MAX_DELTA_MOVEMENT 10
```
natinusala, line 25: Can you add some comments to explain what each of these do please?

natinusala (changes requested):
Here are some more tiny comments, I think I'm starting to get through the review!

Once I go through the whole code I'll try it once again on PC and on Switch to see how it looks and feels, and see if there are little things that need to be tweaked here and there.

natinusala:
TODO for me after I'm done reviewing the code:
- Remember to remove all logs
- See if everything behaves likes HOS
- See if sounds behave like HOS
- Think of how to easily integrate the "tap" gesture to the "click" action we already have, so that it just works (tm)

library/include/borealis/core/touch/scroll_gesture.hpp
XITRIX: I think while it's temporary solution there is no reason to create separate variable for that.
When mouse will be separated from touch, scroll will not trigger by mouse dragging the view.

library/lib/core/application.cpp
XITRIX: I think it will be used also for mouse , so may be leave it as is?

library/lib/core/application.cpp
XITRIX: Could you hint me what else sounds have randomized pitch? It was the only sound I heard randomized.

PR #80, opened 2021-02-15
Fixed scrollviews scrolling when not in focus

HamletDuFromage:
When using a tabFrame, the scrollview of the right panel would scroll with a value calculated with the Y of the focused tab, this makes sure scrollviews only scroll if they're in focus.

I only added these viewtypes because they're the only ones I needed, but this can be extended
```C++
enum class ViewType
{
  VIEW,
  SIDEBARITEM,
};
```

### DIFF
--- a/library/include/borealis/sidebar.hpp
+++ b/library/include/borealis/sidebar.hpp
@@ -47,6 +47,8 @@ class SidebarItem : public View
     Sidebar* sidebar     = nullptr;
     View* associatedView = nullptr;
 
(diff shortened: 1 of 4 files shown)


natinusala:
Fixed in yoga via https://github.com/natinusala/borealis/commit/a7ed4576a1057e2c9b29df815f3a8620a50ed537

natinusala:
Closed as this PR is reimplemented or made irrelevant by the changes on the `yoga` branch.

PR #84, opened 2021-02-26
Adding setText and getText function to Brls::Button

EmreTech:
This Pull Request is for the yoga branch, not the master branch.

In short, make a function in the Brls::button class to change the text of the button.

### DIFF
--- a/demo/components_tab.cpp
+++ b/demo/components_tab.cpp
@@ -33,12 +33,6 @@ ComponentsTab::ComponentsTab()
 bool ComponentsTab::onPrimaryButtonClicked(brls::View* view)
 {
     // TODO: actually do something here
-    primary_button->setText("Some basic text");
-
-    if (primary_button->getText() == "Some basic text") brls::Logger::info("Test was successfull");
-
-    primary_button->setText("Primary button");
-
     brls::Logger::info("Clicked");
     return true;
 }
--- a/demo/components_tab.cpp
+++ b/demo/components_tab.cpp
@@ -33,6 +33,12 @@ ComponentsTab::ComponentsTab()
 bool ComponentsTab::onPrimaryButtonClicked(brls::View* view)
 {
     // TODO: actually do something here
+    primary_button->setText("Some basic text");
+
+    if (primary_button->getText() == "Some basic text") brls::Logger::info("Test was successfull");
+
+    primary_button->setText("Primary button");
+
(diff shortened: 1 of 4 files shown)


library/include/borealis/views/button.hpp
```diff
@@ -163,6 +163,8 @@ class Button : public Box
      */
     void setState(ButtonState state);
 
+    void setText(std::string text);
```
natinusala, line 166: Can you add a documentation comment please ?

natinusala (changes requested):

library/include/borealis/views/button.hpp
EmreTech: Alright, done

EmreTech (commented):
Don't know if you got my comment yet, but I added a documentation comment

demo/components_tab.cpp
```diff
@@ -33,6 +33,12 @@ ComponentsTab::ComponentsTab()
 bool ComponentsTab::onPrimaryButtonClicked(brls::View* view)
 {
     // TODO: actually do something here
+    primary_button->setText("Some basic text");
```
natinusala, line 36: What is the purpose of that? To test that setText and getText work?

While I like the intent, the demo is not the place for such a test. Actually... there is no place for such a test, as we currently don't have a testing suite with borealis.

I want one but I didn't find something that suited all my needs (aka hook into the tested app, run it for multiple frames and test what's going on _inside_). I think we would need a custom made one, which takes time and effort.

natinusala (changes requested):

demo/components_tab.cpp
EmreTech: Done

library/include/borealis/views/button.hpp
```diff
@@ -163,6 +163,17 @@ class Button : public Box
...
     void setState(ButtonState state);
 
+
+    /**
+     * Sets the text of the button. 
+     */
+    void setText(std::string text);
+
+    /**
+     * Gets the text of the button and returns it
```
natinusala, line 173: "gets" is a synonym of "return", there is no need to say "get and return" 😛 

"Returns the text of the button" is fine

EmreTech: Fixed that. I probably should pay more attention to my grammar.

PR #85, opened 2021-03-06
Use romfs for audio on Switch when the program is qlaunch

NicholeMattera:
We talked about this on Discord. When Borealis is being used for a qlaunch replacement there is no need to mount qlaunch's romfs on to another mount point. Because of layeredfs we can access the sound file through the romfs mount point.

### DIFF
--- a/library/lib/platforms/switch/switch_audio.cpp
+++ b/library/lib/platforms/switch/switch_audio.cpp
@@ -23,7 +23,8 @@ namespace brls
 
 #define QLAUNCH_PID 0x0100000000001000
 #define QLAUNCH_MOUNT_POINT "qlaunch"
-#define QLAUNCH_BFSAR_PATH "qlaunch:/sound/qlaunch.bfsar"
+#define ROMFS_MOUNT_POINT "romfs"
+#define BFSAR_PATH "/sound/qlaunch.bfsar"
 
 const std::string SOUNDS_MAP[_SOUND_MAX] = {
     "", // SOUND_NONE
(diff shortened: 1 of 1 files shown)


natinusala:
Thanks for the PR!

Is there the need to redefine the whole path? Only the mount point changes, `romfs:` or `qlaunch`. The path behind stays the same.

Wouldn't it be better to define once and use the same "path" and just change the mount point "prefix" depending on tid?

NicholeMattera:
Updated it with the changes you recommended.

library/lib/platforms/switch/switch_audio.cpp
```diff
@@ -23,7 +23,8 @@ namespace brls
 
 #define QLAUNCH_PID 0x0100000000001000
 #define QLAUNCH_MOUNT_POINT "qlaunch"
-#define QLAUNCH_BFSAR_PATH "qlaunch:/sound/qlaunch.bfsar"
+#define ROMFS_MOUNT_POINT "romfs"
+#define BFSAR_PATH ":/sound/qlaunch.bfsar"
```
natinusala, line 27: I don't like having the `:` in the "path" because technically it's not a valid character for POSIX paths, which we are kinda trying to stay compliant with. I would rather have the `:` added in the sprintf like `%s:%s`.

library/lib/platforms/switch/switch_audio.cpp
```diff
@@ -51,18 +52,28 @@ SwitchAudioPlayer::SwitchAudioPlayer()
...
+        Result result = romfsMountDataStorageFromProgram(QLAUNCH_PID, QLAUNCH_MOUNT_POINT);
+        if (!R_SUCCEEDED(result))
+        {
+            Logger::error("Unable to mount qlaunch ROMFS: {:#x}", result);
+
+            plsrPlayerExit();
+            return;
+        }
+
+        sprintf(bfsarPath, "%s%s", QLAUNCH_MOUNT_POINT, BFSAR_PATH);
```
natinusala, line 70: Does the compiler warn if we add more characters to either defines, making the sprintf overflow the allocated 29 chars? If it doesn't warn, can you then change for snprintf, which does warn?

NicholeMattera: :+1:  Yep it does:
```
/home/nichole/Projects/borealis/./library/lib/platforms/switch/switch_audio.cpp: In constructor 'brls::SwitchAudioPlayer::SwitchAudioPlayer()':
/home/nichole/Projects/borealis/./library/lib/platforms/switch/switch_audio.cpp:73:32: warning: '%s' directive writing 28 bytes into a region of size 23 [-Wformat-overflow=]
   73 |         sprintf(bfsarPath, "%s:%s", ROMFS_MOUNT_POINT, BFSAR_PATH);
      |                                ^~
/home/nichole/Projects/borealis/./library/lib/platforms/switch/switch_audio.cpp:73:16: note: 'sprintf' output 35 bytes into a destination of size 29
   73 |         sprintf(bfsarPath, "%s:%s", ROMFS_MOUNT_POINT, BFSAR_PATH);
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/nichole/Projects/borealis/./library/lib/platforms/switch/switch_audio.cpp:70:32: warning: '%s' directive writing 28 bytes into a region of size 21 [-Wformat-overflow=]
   70 |         sprintf(bfsarPath, "%s:%s", QLAUNCH_MOUNT_POINT, BFSAR_PATH);
      |                                ^~
/home/nichole/Projects/borealis/./library/lib/platforms/switch/switch_audio.cpp:70:16: note: 'sprintf' output 37 bytes into a destination of size 29
   70 |         sprintf(bfsarPath, "%s:%s", QLAUNCH_MOUNT_POINT, BFSAR_PATH);
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

library/lib/platforms/switch/switch_audio.cpp
NicholeMattera: Fixed. :+1:

PR #86, opened 2021-03-06
Added option to enable or disable global actions

NicholeMattera:
Another thing we talked about on Discord. When running as qlaunch you don't want the user from being able to ever exit out of it. The other use case was for updaters (Or even when just an app is updating itself) where if the user were to exit it would leave their SD card or the app itself in a possible broken state. I also included an option to enable and disable the FPS toggle as I could see that being useful for debugging, but probably don't want to keep that in for release builds.

### DIFF
--- a/library/include/borealis/core/application.hpp
+++ b/library/include/borealis/core/application.hpp
@@ -156,7 +156,14 @@ class Application
 
     inline static float windowScale;
 
+    /**
+     * Sets whether BUTTON_START will globally be used to close the application.
+     */
     static void setGlobalQuit(bool enabled);
+
+    /**
+     * Sets whether BUTTON_BACK will globally be used to toggle an FPS display.
+     */
     static void setGlobalFPSToggle(bool enabled);
 
     static GenericEvent* getGlobalFocusChangeEvent();
--- a/library/include/borealis/core/application.hpp
+++ b/library/include/borealis/core/application.hpp
@@ -156,6 +156,9 @@ class Application
 
     inline static float windowScale;
 
+    static void setGlobalQuit(bool enabled);
+    static void setGlobalFPSToggle(bool enabled);
+
     static GenericEvent* getGlobalFocusChangeEvent();
     static VoidEvent* getGlobalHintsUpdateEvent();
 
@@ -202,6 +205,9 @@ class Application
     inline static unsigned blockInputsTokens = 0; // any value > 0 means inputs are blocked
 
     inline static std::string commonFooter = "";
+
+    inline static bool globalQuitEnabled = true;
+    inline static bool globalFPSToggleEnabled = true;
 
     inline static View* repetitionOldFocus = nullptr;
 
--- a/library/include/borealis/core/application.hpp
+++ b/library/include/borealis/core/application.hpp
@@ -28,10 +28,8 @@
 #include <borealis/core/platform.hpp>
 #include <borealis/core/style.hpp>
 #include <borealis/core/theme.hpp>
-
 #include <borealis/core/view.hpp>
 #include <borealis/views/label.hpp>
-
 #include <unordered_map>
 #include <vector>
 
@@ -213,10 +211,10 @@ class Application
 
     inline static std::string commonFooter = "";
 
-    inline static bool globalQuitEnabled = false;
-    inline static int gloablQuitIdentifier = -1;
-    inline static bool globalFPSToggleEnabled = false;
-    inline static int gloablFPSToggleIdentifier = -1;
+    inline static bool globalQuitEnabled                     = false;
+    inline static ActionIdentifier gloablQuitIdentifier      = ACTION_NONE;
+    inline static bool globalFPSToggleEnabled                = false;
+    inline static ActionIdentifier gloablFPSToggleIdentifier = ACTION_NONE;
 
     inline static View* repetitionOldFocus = nullptr;
 
--- a/library/include/borealis/core/application.hpp
+++ b/library/include/borealis/core/application.hpp
@@ -239,6 +239,8 @@ class Application
     static bool handleAction(char button);
 
     static void registerBuiltInXMLViews();
+
+    static ActionIdentifier registerFPSToggleAction(Activity* activity);
 };
 
 } // namespace brls
--- a/library/include/borealis/core/application.hpp
+++ b/library/include/borealis/core/application.hpp
@@ -206,8 +206,10 @@ class Application
 
     inline static std::string commonFooter = "";
 
-    inline static bool globalQuitEnabled = true;
-    inline static bool globalFPSToggleEnabled = true;
+    inline static bool globalQuitEnabled = false;
+    inline static int gloablQuitIdentifier = -1;
+    inline static bool globalFPSToggleEnabled = false;
+    inline static int gloablFPSToggleIdentifier = -1;
 
     inline static View* repetitionOldFocus = nullptr;
 
--- a/library/include/borealis/core/view.hpp
+++ b/library/include/borealis/core/view.hpp
@@ -988,6 +988,11 @@ class View
     void registerAction(std::string hintText, enum ControllerButton button, ActionListener actionListener, bool hidden = false, enum Sound sound = SOUND_NONE);
 
     /**
+     * Unregisters an action for the given button.
+     */
+    void unregisterAction(std::string hintText, enum ControllerButton button, bool hidden = false, enum Sound sound = SOUND_NONE);
+
+    /**
      * Shortcut to register a generic "A OK" click action.
      */
     void registerClickAction(ActionListener actionListener);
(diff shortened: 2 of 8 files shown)


natinusala:
Thanks for the PR!

IMO the global quit option should be `false` by default. I expect borealis to be used to make "proper" apps with a navigation flow between multiple activities, etc... I feel like this global quit option would be a nuisance to be enabled by default.

I have a few requests:
- can you add documentation above the two methods you added? to explain what they toggle, and most importantly, say what key is used for each global action (using the brls controller buttons enum entries)
- can you add a method to "add an exit action" on an activity? that way people can easily add a "+ to exit" feature to their "main menus" if they want to, by calling that method on the activity
- can you make it so that unregistering an action only takes the controller button as parameter? unless we want to support multiple actions on the same view with the same button, but I don't think that would be a good thing to have

A cleaner but harder to implement alternative would be to return an identifier to an action when registering it. That way you can store the identifier and use it later in unregisterAction. It's how Event works, when you subscribe to an event it gives you a "subscription" handle that you can reuse later to remove that one particular callback from the event.

natinusala:
And so the point of all those changes is to remove all the redundancy of the registerAction / unregisterAction calls that give the same hint, button, callback, etc...

NicholeMattera:
> can you make it so that unregistering an action only takes the controller button as parameter? unless we want to support multiple actions on the same view with the same button, but I don't think that would be a good thing to have

I originally had this, but I was afraid of someone overriding the action and then their custom one is wiped out when using `setGlobalQuit(false)`. I like the idea of returning an identifier and using that to unregister an action.

natinusala:
The other solution is to remove the setGlobalQuit() entirely and just require people to call addQuitAction() on every activity if they need to 🤷 I don't really know if people expect + to quit everywhere or just on the "main menu"

Slluxx:
> I don't really know if people expect + to quit everywhere or just on the "main menu"

Imo it would me mildly annoying to have to go to the main menu to quit the application. We are not just used to it in Borealis but the entire switch with the home button. Though it does not quit the application, it does take you home wherever you are on the system. I am certain that changing this "flow" would feel pretty disrupting.

natinusala:
Yeah but take setup wizards for instance, you don't want the user to abort 
the install process in the middle of it

You also may want to save what has changed before quitting. Or put a dialog 
"you have unsaved changes"... etc

I'd say it depends on the app, and by default I'd prefer to disable it to 
avoid "damage" in case it's forgotten.

NicholeMattera:
I will try to work on this tonight. Sorry been preoccupied with other things. 👍

NicholeMattera:
> can you add a method to "add an exit action" on an activity? that way people can easily add a "+ to exit" feature to their "main menus" if they want to, by calling that method on the activity

Just wanted to verify if you wanted to do this. Sounds like you would prefer if people didn't follow this pattern, and if they really want this then they can either use `brls::Application::setGlobalQuit(true);` or just register an action and call `brls::Application::quit();` manually.

natinusala:
Actually now that I think about it, why can't we have both to please everyone?
- make an `addExitAction` method for people who want exit on some activities, like main menu
- keep `setGlobalQuit` for people who want the ability to quit from everywhere, but setGlobalQuit must call addExitAction to reduce redudancy
- global quit is off by default, so that every developer makes a deliberate choice to implement one of the two options

How does that sound?

NicholeMattera:
Done. Also modified the demo so you can exit out of it. I can always change it to use `registerExitAction` or `setGlobalQuit` if you want. I just made it so BUTTON_B would exit. Should `registerExitAction` maybe take a button as an argument, but default the argument to BUTTON_START?

demo/main_activity.hpp
```diff
@@ -23,4 +23,6 @@ class MainActivity : public brls::Activity
   public:
     // Declare that the content of this activity is the given XML file
     CONTENT_FROM_XML_RES("activity/main.xml");
+
+    void onContentAvailable() override;
```
natinusala, line 27: The demo aims to be used as a reference and learn by example, can you add a comment to explain what it does / why it's here? Like I did for the CONTENT_FROM_XML_RES above

library/include/borealis/core/application.hpp
```diff
@@ -203,6 +213,11 @@ class Application
 
     inline static std::string commonFooter = "";
 
+    inline static bool globalQuitEnabled = false;
```
natinusala, line 216: Can you run clang-format please? `./scripts/format.sh --fix`

library/include/borealis/core/view.hpp
```diff
@@ -984,8 +984,16 @@ class View
      * The sound will only be played if the listener returned true.
      *
      * A hidden action will not show up in the bottom-right hints.
+     *
+     * Returns the identifier for the action, so it can be unregistered later on. Returns -1 if the
+     * action was not registered.
+     */
+    int registerAction(std::string hintText, enum ControllerButton button, ActionListener actionListener, bool hidden = false, enum Sound sound = SOUND_NONE);
```
natinusala, line 991: The identifier type should be hidden away from the user, with a typedef and constant for `ACTION_NONE`

`typedef int ActionIdentifier`
`#define ACTION_NONE -1`

library/lib/core/application.cpp
```diff
@@ -435,6 +433,27 @@ void Application::toggleFramerateDisplay()
...
+            (*it)->unregisterAction(Application::gloablQuitIdentifier);
+    }
+}
+
+void Application::setGlobalFPSToggle(bool enabled) {
+    Application::globalFPSToggleEnabled = enabled;
+    for (auto it = Application::activitiesStack.begin(); it != Application::activitiesStack.end(); ++it) {
+        if (enabled)
+            Application::gloablFPSToggleIdentifier = (*it)->registerAction(
+                "FPS", BUTTON_BACK, [](View* view) { Application::toggleFramerateDisplay(); return true; }, true);
```
natinusala, line 451: It looks like I missed this, sorry, the FPS toggle should also be in its own (private?) method to avoid redudancy between that and line 565

natinusala (changes requested):

demo/main_activity.hpp
NicholeMattera: Done. :+1:

library/include/borealis/core/application.hpp
NicholeMattera: I had to modify the script on my computer to run as I have clang 11 installed on my computer, and didn't see a way to easily install clang 8 on Arch. I did run it though. (I only committed the files that I've touched in this PR.) :+1:

library/include/borealis/core/view.hpp
NicholeMattera: Done. :+1:

library/lib/core/application.cpp
NicholeMattera: Done. :+1:

library/lib/core/activity.cpp
```diff
@@ -112,10 +115,26 @@ bool Activity::isHidden()
...
-void Activity::registerAction(std::string hintText, enum ControllerButton button, ActionListener actionListener, bool hidden, enum Sound sound)
+ActionIdentifier Activity::registerAction(std::string hintText, enum ControllerButton button, ActionListener actionListener, bool hidden, enum Sound sound)
+{
+    if (this->contentView)
+        return this->contentView->registerAction(hintText, button, actionListener, hidden, sound);
+
+    return ACTION_NONE;
+}
+
+void Activity::unregisterAction(int identifier)
```
natinusala, line 126: `int` 😛

natinusala (changes requested):

library/lib/core/activity.cpp
NicholeMattera: Sorry, it's taken me so long to fix this. I've had a hell of a week thus far. 😭

natinusala:
Actually :see_no_evil: After testing your changes I realized that I use START to quit the demo myself when I'm writing code for the library. I open the demo, go where I want to test my stuff, test it and press START to quit.

So... can you enable global quit for the demo please?

NicholeMattera:
> Actually see_no_evil After testing your changes I realized that I use START to quit the demo myself when I'm writing code for the library. I open the demo, go where I want to test my stuff, test it and press START to quit.
> 
> So... can you enable global quit for the demo please?

Changed it to use the global quit and added documentation.

NicholeMattera:
Something I thought about... Would it make more sense to change `std::vector<Action> actions;` to be `std::unordered_map<ControllerButton, Action> actions;`. Actions are unique by ControllerButton and this should come with a slight performance increase. I can wait and do a separate PR if you want. It would just touch a lot of the same files.

natinusala:
Well if you can make it so that the original behavior is entirely kept, sure. That would be in another PR though :p

PR #87, opened 2021-03-07
Make the devide by 0 button actually devide by 0

Tim-Hobbes:
Currently when you press the divide by 0 there is no actual division taking place: there is just a call to an artificial crash. I fixed it so there is an actual division taking place and the answer is displayed in a notification just in case it succeeds.

### DIFF
--- a/example/main.cpp
+++ b/example/main.cpp
@@ -103,8 +103,11 @@ int main(int argc, char* argv[])
             "main/jank/ios"_i18n,
         });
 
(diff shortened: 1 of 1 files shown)


natinusala:
Is... Is this a serious submission?

The button is to showcase the fake "crash" dialog

Tim-Hobbes:
Kinda serous, would you like me to make PozzNX also functional :-) ? If it was to showcase the error popup I will close the pull request.

natinusala:
The new demo will have actual showcases with meaningful buttons instead of memes like the current example.

Issue #140, opened 2021-03-09
Storage system


natinusala:
Storage can be used to store anything on disk between app "sessions". The aim is not for the user to configure the app, so the files don't have to be human readable and most importantly the files don't need to contain every entry with their default value, like a config file would.

This is different from a "config" engine, which purpose is for the user to configure the app. Here, the storage is used for the app to maintain its state between runs, including but not limited to config. For instance, a "has the setup wizard been executed yet" boolean can be kept in the storage, but it's not semantically a config field. A list of favorited entries also belongs in the storage, and can be considered configuration, but it doesn't matter.

A storage file is in XML, to avoid adding yet another language library (sorry TOML).

Each platform implementation will give the path to where the storage files are to be saved and read. On Switch it can be `/config/app-name` even though it's not _technically_ config, we'll keep the standard.

For the API, I want the entries to be typed and checked at compile time, in a declarative manner (so header only). No code should be required to get it going. We can use macros for that.

For instance the following code (types such as bool, string, string list... and how they are defined are only for example purposes):

```c++
#include <borealis.hpp>

class SettingsStorage : public brls::StorageFile
{
  BRLS_STORAGE_FILE_INIT("settings");
  
  BRLS_STORAGE_BOOL(wizardShown, "wizard_shown");

  BRLS_STORAGE_STRING(username, "username");
  BRLS_STORAGE_STRING(password, "password");
  
  BRLS_STORAGE_STRING_LIST(favorites, "favorites");
  
  BRLS_STORAGE_STRING_INT_MAP(launchCount, "launch_count");
}
```

Would be used this way:

```c++
SettingsStorage* settings = SettingsStorage::open();

bool wizardShown = settings->wizardShown; // cast operator overload

for (std::string favorite : settings->favorites)
	brls::Logger::info("{}", favorite);
	
settings->username = "h4ck3rm4n"; // assignment operator overload
settings->username.save(); // to save only one property

settings->favorites.push_back("borealis");

settings->save(); // to save all edited properties
```

And would generate the following XML in `/config/app-name/settings.xml`:

```xml
<brls:StorageFile>
	<brls:BoolProperty name="wizard_shown" value="true" />
	
	<brls:BoolProperty name="username" value="h4ck3rm4n" />
	
	<brls:StringListProperty name="favorites">
		<brls:Value value="borealis" />
	</brls:StringListProperty>
</brls>
```

This is of course just a draft / example of how I would like it to be. I'm sure typing can be handled better than that.

[...]

PR #146, opened 2021-03-09
Core - StorageFile

EmreTech:
Check #140 for context clues

This is my implementation of the StorageFile class idea, for the release of Borealis. This is a work-in-progress Pull Request since I just started on this.

To-Do with this Pull Request:
- Nothing for now!
- Stop committing changes every 5 minutes and actually test a commit before committing (suggestion from natinusala)

What's finished:
- [x] Fix bug where values from ListStorageObject are added twice when reading from file
- [x] Re-add ListStorageObject
- [x] Rewrite the code to fit the original design more


EmreTech:
Okay, sounds good!

EmreTech:
I didn't do a compilation test until now, and there were some compiler errors. So I went ahead and fixed them up.

EmreTech:
The class in its simplest, probably functional form is finished. I'm open for a review now.

EmreTech:
There are a lot of issues with the class (i.e. the writing and reading functions not working correctly) so I'm going to try and fix them.

EmreTech:
So I have a new idea. Instead of using my approach from yesterday, I can use a new one that should bring this implementation of this idea closer to the original idea. There will be a new class called StorageObject, which contains variables for the value, name, and type (like int or string for example).

 In StorageFile, there will be a whole std::vector containing these StorageObjects. This means readFromFile can read from the std::vector, not directly from the XML file. writeToFile should be the same, and there will be a new function to parse the XML file into StorageObject objects.

I was also thinking of implementing the save function with individual elements, but that might require more work.

natinusala:
I'm pretty sure what I wrote in the issue can be implemented exactly as I wrote it, it's a design more than an "idea"

EmreTech:
Yeah, I think I can do it if I put more effort. Right now, you don't have to write a lot of code, but you might still have to write code.

EmreTech:
I just confirmed, and now everything works as intended! I had to fix up some runtime errors mostly relating to how I was treating pointers. (I was treating them more like a normal variable than a pointer, there were many Segmentation Faults)

EmreTech:
Wow, I messed up. Tried to get the latest changes from the upstream main branch.

EmreTech:
Actually wait, I didn't, but it's a big mess

natinusala:
image

What did you do 👀

EmreTech:
I did a `git rebase` instead of `git merge`...

I wanted to update my fork since there was a new commit.

EmreTech:
Yes, I literally don't know how to properly use git. I use GitHub Desktop most of the time

natinusala:
Rebase is the right choice when making a PR

EmreTech:
Yeah, but I'm not sure why it recommitted everything that I already committed to this Pull Request

natinusala:
Because when you rebase it removes all your commits, updates the branch and re-applies your commits back on the updated branch. As a consequence, your commits are not the same as the previous ones (same content but new hash). The history has also been rewritten, making a force push mandatory. So since the commits are all same same, but different, but still same, GitHub treats them all as new commits in the PR.

EmreTech:
Actually, yes, I should've done a force push. I pushed directly through GitHub Desktop, so it must've not done a force push. I was thinking of reverting to the previous commit before I rebased, and try again.

EmreTech:
Well, it happened again. :(

EmreTech:
Well actually, I fixed it! But it counted all the commits from today. And it didn't update from the upstream

EmreTech:
Alright, I have finally tested on the Nintendo Switch! It seems to work fine, just the weird segmentation faults sometimes.

EmmmaTech:
So because your issue was a "design", not an "idea", I decided to rewrite the entire PR to match the design a lot more. It has some missing functions from the original design, but otherwise, it works fine.

EmmmaTech:
Reopened in #175 due to branch reorganization.

PR #153, opened 2021-03-20
Recycler view

XITRIX:
I'm in progress of creating recycler view.

I've changed the behaviour of detached views a bit. If previously they were fully detached from parent's coordinate system, now they still attached to parent's origin, but free from Yoga's layout calculations, so detached origin can be used to modify view's local position inside it's parent. I changed it cause I need cells to be attached to scroll's contentBox, so it's scrolling position will reflect to cell's absolute position, but still need to have an ability to place cells manually. Also I can't think of why we could need to absolutely detach view from it's parent, I think the behaviour I've made could be much more useful than previous one. If you have any idea about that, I'll be glad to hear them. 

That PR is still in progress, but I'd like you to know about it.


natinusala:
Hey thanks a lot for that draft!

You are right about the detached views changed, I should have done it like that since the beginning.

Since this starts from your work on touch, I would prefer to finish reviewing and merging that first, then I'll come to this PR, does that sound good to you?

XITRIX:
Yes, I'm also thought about that, there is no need in reviewing same changes twice. Let's finish with touch, then come back here.

EmreTech:
I was honestly thinking about that. It doesn't make sense to have a PR that uses code from another PR that hasn't been merged yet.

XITRIX:
True, but it uses a lot of Point and Rect structs, which I added in Touch PR so I decided to do it like that.

Issue #130, opened 2021-03-09
Properly build file paths


natinusala:
File paths are created using string concatenation with hardcoded `/` path separator. This is bad. Paths should be platform agnostic and build properly instead of hardcoded like that.

It may be a good occasion to rework / replace the `BRLS_ASSET` macro, that maybe needs to be more than a macro.

Examples:
- i18n loadLocale
- View createFromXMLResource

Issue #127, opened 2021-03-09
i18n improvements


natinusala:
- [ ] Switch to XML to remove JSON dependency and reduce binary bloat
- [ ] Remove the linter (F)
- [ ] Add support for string arrays (ala Android)
- [ ] Move ALL built-in brls strings to the source code instead of resources files (like built-in views XML)
- [ ] Add an opt-out feature to entirely disable i18n - when doing so, specify a new default language to use for built-in views (since they will still be using i18n)

PR #157, opened 2021-03-27
Core - Improvements to i18n

EmmmaTech:
Implements #127 and #130, all in one Pull Request.
This Pull Request aims to improves i18n and BRLS_ASSET.

Features implemented before this PR was made:
- Switches from using JSON to XML for i18n
- Removes the linter script (partly replaced with printing errors)
- Move built-in brls strings to source code (like built-in views XML)
- Add an opt-out feature to disable i18n (still loads internal translations)
- Change BRLS_ASSET from a macro to an inline static function
-  Use std::filesystem::path to convert to a platform-specific directory separator
- Add method to get locale for the GLFW Platform (tested on macOS Big Sur 11.0.1)

To-do with this Pull Request:
- Nothing, for now

What's finished:
- See "Features implemented before this PR was made"
- [x] Write a Python script to convert an i18n JSON to i18n XML
- [x] Fully re-implement linter script inside of the library
- [x] Change how internal translations are loaded to what it was before (instead of loading to a separate unordered map, load to the current or default locale unordered map)


EmreTech:
I like how I replaced a fatty library with a even more fatty library

EmreTech:
No, I downloaded the 2021.03.22.00 release, It should be a subrepo

EmreTech:
Okay done. It's now a subrepo! Also, just so you can know, the Switch build still uses `std::unordered_map`. IIRC, I couldn't get it to build, so GLFW will be using F14FastMap for now.

natinusala:
Couldn't get what to build? F14FastMap?

EmreTech:
Yeah. It freaks out after it can't find file `folly-config.h`. And even after providing that file, it still didn't work IIRC.

EmmmaTech:
I decided to remove Folly because the library is too big and we're only using it for one purpose. Someone else can add it back later.

PR #159, opened 2021-04-02
add Image::setImageFromMem

FaultyPine:
Allows images to be set from raw data

### DIFF
--- a/library/lib/views/image.cpp
+++ b/library/lib/views/image.cpp
@@ -260,6 +260,28 @@ void Image::setImageFromFile(std::string path)
     this->invalidate();
 }
 
+void Image::setImageFromMem(unsigned char* data, int size)
+{
+    NVGcontext* vg = Application::getNVGContext();
+
+    // Free the old texture if necessary
+    if (this->texture != 0)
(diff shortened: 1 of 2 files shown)


library/lib/views/image.cpp
```diff
@@ -260,6 +260,28 @@ void Image::setImageFromFile(std::string path)
...
+
+    // Free the old texture if necessary
+    if (this->texture != 0)
+        nvgDeleteImage(vg, this->texture);
+
+    // Load the new texture
+    this->texture = nvgCreateImageMem(vg, 0, data, size);
+
+    if (this->texture == 0)
+        fatal("Cannot load image from mem!");
```
natinusala, line 275: no need for an exclamation mark :p

natinusala (changes requested):
Thanks for the PR! It looks good to me

Do you think you could add a way to test / showcase this in the demo?

EmmmaTech:
As natinusala said, you should really add an demonstration of this in the demo

EmmmaTech:
Otherwise, no other problems

PR #162, opened 2021-04-29
Added Image::setImageFromMemory and Image::setImageFromRawData

iUltimateLP:
Adds support to set the image resource a brls::Image view shows from memory using nanovg's nvgCreateImageMem.

### DIFF
--- a/demo/dynamic_image.cpp
+++ b/demo/dynamic_image.cpp
@@ -59,7 +59,7 @@ bool DynamicImage::onImageClicked(brls::View* view)
     }
 
     // Display the image
-    this->image->setImageFromMemory(imageData, width, height);
+    this->image->setImageFromRawData(imageData, width, height);
 
     // Free the image buffer
     free(imageData);
--- a/demo/dynamic_image.cpp
+++ b/demo/dynamic_image.cpp
@@ -1,0 +1,74 @@
+/*
+    Copyright 2019 WerWolv
+    Copyright 2019 p-sam
+    Copyright 2020-2021 natinusala
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include "dynamic_image.hpp"
+
+DynamicImage::DynamicImage()
+{
+    // Load the XML file and inflate ourself with its content
+    this->inflateFromXMLRes("xml/views/dynamic_image.xml");
+
+    // Forward Image XML attributes
+    this->forwardXMLAttribute("scalingType", this->image);
+    this->forwardXMLAttribute("image", this->image);
+    this->forwardXMLAttribute("focusUp", this->image);
+    this->forwardXMLAttribute("focusRight", this->image);
+    this->forwardXMLAttribute("focusDown", this->image);
+    this->forwardXMLAttribute("focusLeft", this->image);
+    this->forwardXMLAttribute("imageWidth", this->image, "width");
+    this->forwardXMLAttribute("imageHeight", this->image, "height");
+
+    // Register a click action for this button
+    BRLS_REGISTER_CLICK_BY_ID("image", this->onImageClicked);
+}
+
+bool DynamicImage::onImageClicked(brls::View* view)
+{
+    // Retrieve the image size
+    int height = this->image->getHeight();
+    int width = this->image->getWidth();
+
+    // Generate the pixel data for a new image
+    size_t bufferSize = height * width * 4; // 4 bytes per pixel (RGBA)
+    unsigned char* imageData = (unsigned char*)calloc(1, bufferSize);
+
+    for (int y = 0; y < height; y++)
+    {
+        for (int x = 0; x < width; x++)
+        {
+            memset(&imageData[(y * width + x) * 4 + 3], (unsigned char)255, sizeof(unsigned char));                         // Alpha
+            memset(&imageData[(y * width + x) * 4 + 2], (unsigned char)((y+x)*255/(height+width)), sizeof(unsigned char));  // Blue
+            memset(&imageData[(y * width + x) * 4 + 1], (unsigned char)(x*255/width), sizeof(unsigned char));               // Green
+            memset(&imageData[(y * width + x) * 4 + 0], (unsigned char)(y*255/height), sizeof(unsigned char));              // Red
+        }
+    }
+
+    // Display the image
+    this->image->setImageFromMemory(imageData, width, height);
+
+    // Free the image buffer
+    free(imageData);
+
+    return true;
+}
+
+brls::View* DynamicImage::create()
+{
+    // Called by the XML engine to create a new DynamicImage
+    return new DynamicImage();
+}
--- a/library/include/borealis/views/image.hpp
+++ b/library/include/borealis/views/image.hpp
@@ -96,6 +96,15 @@ class Image : public View
     void setImageFromFile(std::string path);
 
     /**
+     * Sets the image from a memory buffer. The data read has to be
+     * the same as if the image would've been read from a file.
+     * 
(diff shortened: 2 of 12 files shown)


library/lib/views/image.cpp
```diff
@@ -280,6 +281,34 @@ void Image::setImageFromFile(std::string path)
...
+    // Load the new texture
+    int flags     = this->getImageFlags();
+    this->texture = nvgCreateImageMem(vg, flags, data, numData);
+ 
+    if (this->texture == 0)
+    {    
+        std::stringstream addr;
+        addr << (void*)data;
+
+        fatal("Cannot load image from memory (" + std::to_string(numData) + " bytes at 0x" + addr.str() + ")");
```
natinusala: Why do you need a stringstream for that?

natinusala:
Hi, thanks for the PR!

This is a duplicate of #159 but I can merge whichever PR is finished first.

Do you think you could add a way to test / showcase this in the demo?

library/lib/views/image.cpp
iUltimateLP: To convert the address of the pointer to a std:: string (it needs to be a std::string so I can use the operator+). If there's an easier way though, let me know!

iUltimateLP:
Ah didn't see the other PR :)

As for a demo case, I'd implement a new view element "DynamicImage" or something which generates a random Bitmap and renders that!

library/lib/views/image.cpp
natinusala: I don't think that's worth just for an error message IMO

iUltimateLP:
Had some nasty bugs with memory allocation, which is why I switched from `nvgCreateImageMem` (which was pretty unnessecary anyway since it needed a whole image "file" in memory rather than just a raw rgba array) to `nvgCreateImageRGBA`. Also implemented a little example in `demo/dynamic_image` and `views/dynamicImage.xml`. If you click on the image, it will populate with a gradient - let me know what you think!

natinusala:
This is absolutely not the same, and if I were to choose I would keep the "from mem" instead of "from RGBA" method.

The "from mem" method is used to load JPG / PNG files from memory, and is actually currently used in the sys-clk manager to display game thumbnails. Any homebrew dealing with images on HOS would need it too (homebrew menu, qlaunch replacement, album replacement...).

I understand that it's hard to generate a random PNG in memory, so how about adding both methods and only making an example for thr RGBA one ?

iUltimateLP:
See my latest commit, I had the same thought :P

demo/dynamic_image.cpp
```diff
@@ -0,0 +1,74 @@
+/*
+    Copyright 2019 WerWolv
```
natinusala: are you sure of the copyright ?

demo/dynamic_image.cpp
```diff
@@ -0,0 +1,74 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include "dynamic_image.hpp"
+
+DynamicImage::DynamicImage()
```
natinusala: I would not call it "dynamic image", I don't find it very explicit

How about something like "RandomNoiseImage"? Or RandomRGBAImage?

library/include/borealis/views/image.hpp
```diff
@@ -95,6 +95,23 @@ class Image : public View
...
+     */
+    void setImageFromMemory(unsigned char* data, int numData);
+
+    /**
+     * Sets the image from a memory buffer. The data should be a
+     * raw four-channel (RGBA) pixel array of the size width*height*4.
+     * 
+     * See the example implementation dynamic_image in the demo.
+     */
+    void setImageFromRawData(const unsigned char* data, int width, int height);
```
natinusala: "raw data" can be anything: RGB, RGBA, YUV, ABRG...

I would call it "setImageFromRGBA" rather than "raw data", and most importantly specify the _full_ pixel format in the documentation (RGBA8888 if I'm not mistaken?). This is better than asking for a "width * height * 4" array.

library/lib/views/image.cpp
```diff
@@ -280,6 +280,53 @@ void Image::setImageFromFile(std::string path)
...
+    if (this->texture != 0)
+        nvgDeleteImage(vg, this->texture);
+
+    // Load the new texture
+    int flags     = this->getImageFlags();
+    this->texture = nvgCreateImageRGBA(vg, width, height, flags, data);
+
+    if (this->texture == 0)
+    {
+        brls::Logger::error("Cannot load image from memory (" + std::to_string(width) + "x" + std::to_string(height) + ")");
```
natinusala: wrong error string

demo/dynamic_image.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // Free the image buffer
+    free(imageData);
+
+    return true;
+}
+
+brls::View* DynamicImage::create()
+{
+    // Called by the XML engine to create a new DynamicImage
+    return new DynamicImage();
```
natinusala: file missing from the demo meson file

resources/xml/views/dynamic_image.xml
```diff
@@ -0,0 +1,20 @@
...
+        width="auto"
+        height="auto"
+        focusable="true"
+        marginBottom="20px" />
+
+    <brls:Label
+        width="auto"
+        height="auto"
+        horizontalAlign="center"
+        text="Click the image!" />
```
natinusala: Please use i18n, and I think it would be better to have "Randomly generated RGB (click to refresh)" as a caption

demo/dynamic_image.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    // Retrieve the image size
+    int height = this->image->getHeight();
+    int width = this->image->getWidth();
+
+    // Generate the pixel data for a new image
+    size_t bufferSize = height * width * 4; // 4 bytes per pixel (RGBA)
+    unsigned char* imageData = (unsigned char*)calloc(1, bufferSize);
+
+    for (int y = 0; y < height; y++)
```
natinusala: Do you think you could randomize it?

And also make it so that it's "populated" when the view is created, so that it doesn't start all white?

natinusala (changes requested):

demo/dynamic_image.cpp
iUltimateLP: Never worked with PRs and licensing, should it just be `Copyright 2020-2021 natinusala`?

natinusala: No, the copyright is for the person who wrote the file so that would be 2021 [your name]!

demo/dynamic_image.cpp
iUltimateLP: Trying to run the image generation in the RandomRGBAImage constructor causes a crash, I'm pretty sure it's because this->image->getHeight() / width() returns 0, hence it tries to allocate 0 bytes. At what stage is it safe to get the dimensions?

natinusala: You don't need the image width or height, you can just generate an image on a 256x256 grid and borealis will upscale it

natinusala:
This is great! I feel like the example image is a little bit too big and too close to the other images (it's missing some padding).

Could you maybe make it a rectangle, like the "stretched" image example? And add some top padding. Thanks!

iUltimateLP:
Sure, here you go :)

iUltimateLP:
Any update @natinusala? :)

library/lib/views/image.cpp
```diff
@@ -280,6 +280,53 @@ void Image::setImageFromFile(std::string path)
...
+
+    int width, height;
+    nvgImageSize(vg, this->texture, &width, &height);
+    this->originalImageWidth  = (float)width;
+    this->originalImageHeight = (float)height;
+
+    this->invalidate();   
+}
+
+void Image::setImageFromRGBA(const unsigned char* data, int width, int height)
```
EmmmaTech, line 306: Why is this function only `const unsigned char *` and the other isn't? Is there a reason for this?

EmmmaTech (changes requested):
Since natinusala is busy, I decided to look at this PR myself. Here is one change you should consider making

library/lib/views/image.cpp
iUltimateLP: Absolutely right, can definitely be made const.

PR #164, opened 2021-05-21
swkbd: fix support for latest libnx commits.

DarkMatterCore:


natinusala:
Thanks for the PR! I'll merge it when the libnx changes get released.

PR #166, opened 2021-06-08
legacy: translate axis events to D-Pad events.

DarkMatterCore:
I dunno if this feature is actually unwanted for some reason, but I needed to implement it while using the `legacy` branch. Hope it helps.

Please let me know if there's anything wrong so I can fix it.


DarkMatterCore:
I'm switching to `main`.

PR #167, opened 2021-06-17
legacy: ScrollView: fix content view start scroll offset when the sidebar item is initially focused.

DarkMatterCore:
Previous behaviour would mistakenly use boundary values from the focused `SidebarItem` *instead* of a child from the `ScrollView`'s content view, leading to content views not being initially rendered at scroll offset `0.0f`.

This issue becomes evident when using long content views for `SidebarItem` elements beyond the first one.

The problem is fixed by checking if the focused view's parent matches the `ScrollView`'s content view. If the focused item isn't a child of the content view, `currentSelectionMiddleOnScreen` is set to the content view's Y coordinate.

### DIFF
--- a/library/lib/scroll_view.cpp
+++ b/library/lib/scroll_view.cpp
@@ -141,8 +141,16 @@ bool ScrollView::updateScrolling(bool animated)
     if (contentHeight == 0)
         return false;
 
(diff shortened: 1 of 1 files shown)


DarkMatterCore:
I'm switching to `main`.

Issue #132, opened 2021-03-09
Rework style / theme


natinusala:
Having to add styles and themes entries programmatically is bad. They need to be loaded externally from XML files, called "stylesheets". Every built-in view should have its own stylesheet for clarity.

A "theme" is now the result of loading all "stylesheet" files.

The need is for users to:
- redefine styles of built-in views
- create their own theme from scratch
- add theming features to their own custom views

At first, it should be a 1:1 of what we have in the code, aka:
1. Declare a theme (for now there is only one "brls/switch" one)
2. For each variant, declare colors and metrics
3. By default, all apps use the "brls/switch" theme unless specified otherwise by the user
4. The API should stay roughly the same to quickly load theme values in the code and in XML files

So for instance, each built-in view's stylesheet can look like this:
```xml
<brls:Stylesheet theme="brls/switch" prefix="brls/sidebar">
    <brls:ThemeVariant name="dark">
        <brls:Color name="background" value="#323232" />
        <brls:Color name="active_item" value="#00ffcc" />
        <brls:Color name="separator" value="#515151" />
        <brls:Color name="some_other_color" value="rgb(255, 255, 255)" />

        <brls:Metric name="border_height" value="16px" />
        <brls:Metric name="padding_top" value="32px" />

        <!-- And so on... -->
    </brls:ThemeVariant>
</brls:Theme>
```

Which would be equivalent to the current C++ theme and style:
```c++
// Theme
{ "brls/sidebar/background", nvgRGB(50, 50, 50) },
{ "brls/sidebar/active_item", nvgRGB(0, 255, 204) },
{ "brls/sidebar/separator", nvgRGB(81, 81, 81) },

// Style
{ "brls/sidebar/border_height", 16.0f },
{ "brls/sidebar/padding_top", 32.0f },
```

Users can redefine the style by overriding values for the "brls/switch" theme. The final "theme" is the aggregation of all stylesheets, first the built-in ones then the user ones on top of it.

So one challenge is how to build a full theme in one go when the app inits, while having multiple stylesheet files on both the filesystem AND scaterred around the code (for built-in views, I don't want them in XML files and I don't want one big theme file either as we currently have).

Rough summary:
- [ ] Declare colors in XML files (current "theme"), from both hex and rgb(a) formats
- [ ] Declare metrics in XML files (current "style")
[...]

PR #174, opened 2021-12-20
Core - Rework style / theme

EmmmaTech:
This one tackles issue #132, where the current Style/Theme system is reworked.

Here's what's done so far:
- Reworked the concept of Style and Theme, where Theme stores both colors/metrics
- Changed from using a programmatically way to define values to an XML way
- Made Style wrapper for backward compatibility

### DIFF
--- a/demo/main.cpp
+++ b/demo/main.cpp
@@ -72,6 +72,29 @@ int main(int argc, char* argv[])
     brls::getStyle().addMetric("about/description_margin", 50);
     */
 
+    const std::string captionedImageThemeXML = R"xml(
+        <brls:Stylesheet theme="brls/default" prefix="captioned_image">
+            <brls:ThemeVariant name="light">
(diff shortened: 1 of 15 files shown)


JSH32:
Ok I know this is a bad place to contact but is your fork of borealis stable enough to use in a normal app? I'm noticing problems with borealis currently like the fact that it was in a rewrite when I cloned it over. Hints don't show and all that. I assume you fixed that based on the issue history of this repo?

EmmmaTech:
It might be. You'll have to merge the commits from my libnx branch fix and my other feature branches in order to use the features I implemented

JSH32:
Does your fork have any way to make a list of various objects, maybe like a game list? Or like the old implemented full width buttons in SimpleSaveManager for example?

EmmmaTech:
You should be able to make that with the tools you're provided with, but I don't think that's built into the yoga rewrite

PR #168, opened 2021-08-11
Core and Platforms - Updated for latest libnx and fixed compile error

EmmmaTech:
This is identical to #164 except an additional compile error was fixed.

PR #177, opened 2023-10-31
Support newer libnx version

DisabledMallis:
It appears that `swkbdConfigSetStringLenMaxExt` was replaced with `swkbdConfigSetStringLenMin` on the libnx repo... I was having trouble building the demo because of this. I also noticed std::optional was used but not included in the dx_renderer.hpp file? Not sure if thats a change that should be kept, but that was also causing problems for me.


EmmmaTech:
This is a duplicate of #168, which fixes the same problem.

DisabledMallis:
Apologies, I had not seen that. Closing

